PR #36, opened 2019-11-05
Update README.md

runaway518:
Build in Mac OS X

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ For the latest versions of mosquitto we need to export some flags before buildin
 
 ```
 export CGO_CFLAGS="-I/usr/local/include -fPIC"
+export CGO_LDFLAGS="-undefined dynamic_lookup" (MacOS)
 export CGO_LDFLAGS="-shared"
 make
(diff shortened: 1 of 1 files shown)


iegomez:
Hey, thanks for the addition, but could you write separate versions for both and add any details? Something like this (not exactly this probably, I don't own a Mac):
### Debian (and maybe others)

This assumes that mosquitto.h, mosquitto_plugin.h and mosquitto_broker.h are located at /usr/local/include, which is true for a manually built mosquitto version in Debian based systems (and probably others too).

```
export CGO_CFLAGS="-I/usr/local/include -fPIC"
export CGO_LDFLAGS="-shared"
make
```
### MacOS

In MacOS... (add any detail here).

```
export CGO_CFLAGS="-I/usr/local/include -fPIC"
export CGO_LDFLAGS="-undefined dynamic_lookup"
export CGO_LDFLAGS="-shared"
make
```

runaway518:
Hey @iegomez, I just update README!

PR #37, opened 2019-11-05
update test tools

sseide:
for easier comparison of pbkdf2 hash i added a new commad line tool using your library functions.
Hope it helps or may be usefull in it self for some shell scripting

* add cli tool to compare password with pbkdf2 hash
* allow setting keylwght on generating new hash with pw cli tool

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.so
 *.a
 pw
+pwcheck
 go-auth.h
 
 # Test binary, build with `go test -c`
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ all:
 	go build -buildmode=c-archive go-auth.go
 	go build -buildmode=c-shared -o go-auth.so
 	go build pw-gen/pw.go
+	go build pw-gen/pwcheck.go
 
 requirements:
 	dep ensure -v
--- a/pw-gen/pw.go
+++ b/pw-gen/pw.go
@@ -7,18 +7,16 @@ import (
 	"github.com/iegomez/mosquitto-go-auth/common"
 )
 
-// saltSize defines the salt size
-const saltSize = 16
-
 func main() {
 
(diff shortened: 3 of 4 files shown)


iegomez:
Hi, thanks for the PR. I like the addition of the user definable `salt` length, but I'd refactor the tool to look something like this:
```go
func main() {

	var password = flag.String("p", "", "plain text password")
	var hashed = flag.String("h", "", "pbkdf2 password hash")

	flag.Parse()

	if common.HashCompare(*password, *hashed) {
		fmt.Println("success: plain and hashed passwords match")
		return
	}
	fmt.Println("error: plain and hashed passwords don't match")
}
```
If you can add those changes and a mention on the README's section for the `pw` tool, I'll merge it.

sseide:
Ok, code updated.

But i'd like to leave a non-zero exit code if password does not match for easier scripting usage. Only have to compare exit code than instead of text parsing....

iegomez:
Hi, sorry for the delay. I think the requirement is a bit too specific to be included with the more general plugin, so I created a separate repo that calls the `common` package here: https://github.com/iegomez/pw-test.

So, if you'd like to remove the test utility part and leave only the ability to change the salt size, I'll merge this. It'd be nice to also include a link to the testing tool and maybe improve a bit the docs regarding the usage of `pw` at this section https://github.com/iegomez/mosquitto-go-auth#files.

Let me know if you'd be interested in adding that.

Cheers!

iegomez:
@sseide Any news on this? I'm asking to manually add the salt size changes if you're not interested.

sseide:
Sorry, to much to do right now.  and have not found time to check for pbkdf2 differences either...  But if you only want to add salt param its problably faster to add this one param manually.

iegomez:
I've just added the `saltSize` flag and comments at the Readme. Thanks again for bringing this up.

Cheers!

PR #40, opened 2019-12-10
Debian packaging support - first release

wildstray:


iegomez:
I'm not sure about patching the file to import `mosquitto_internal.h` and modifying `clientid` and `username`. I believe you're having some local issue: maybe had an earlier version of Mosquitto and there's something that got left there and is reporting the wrong `MOSQ_AUTH_PLUGIN_VERSION` (had that happen to me some time ago), you have mismatched mosquitto and mosquitto-dev versions, you messed up some flag that's needed for building, etc.. I think you should better check that, as those are the correct definitions for Mosquitto 1.5 and over, so the changes shouldn't be necessary.

Also, why 1.5.7? Is that Debian's latest version?

wildstray:
Ok, I have to deal directly with mosquitto_internal.h and bypass accessory functions mosquitto_client_id() and mosquitto_client_username() because in mosquitto-dev package they are not available (probably not compiled in). 1.5.7 is the latest version used in Debian 10.

iegomez:
Hey @wildstray, just checking up on this. Were you able to tell what was the issue?

iegomez:
Hey again, any updates on this?

Issue #22, opened 2019-06-16
HTTP plugin open file leak


wattanakorn495:
the mosquitto log show too many file open

then i try counting open file in /proc/'PID'/fd 
using this command ""watch -n 0.5 "ls -l | wc -l"
and found that when broker get new client connection, opened file increase by 3 (maybe from GetUser, GetSuperuser and CheckAcl ) and when the client disconnect, opened file reduce only 1 

and i try change "return httpRequest..." to "return true"
the problem is solved

it may have some bug in httpRequest function

iegomez:
Thanks for reporting this. I'll give it a look and try to fix it as soon as possible.

iegomez:
I haven't been able to reproduce this. Are you sure it's the plugin the one not returning the file descriptors and not something else? Are you testing this against an HTTP service in the same machine which could potentially be the culprit?

wattanakorn495:
@iegomez 
sorry for late response

i solved it by moving htpp client and http transport into HTTP struct 

```go
type HTTP struct {
	UserUri      string
	SuperuserUri string
	AclUri       string
	Host         string
	Port         string
	WithTLS      bool
	VerifyPeer   bool
	ParamsMode   string
	ResponseMode string

	Client    *h.Client
	Transport *h.Transport
}
```

and initialize inside NewHTTP

```go
	tr := &h.Transport{ 
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
	}
	client := &h.Client{Timeout: 5 * time.Second}
	client.Transport = tr
	//Initialize with defaults
	var http = HTTP{
		WithTLS:      false,
		VerifyPeer:   false,
		ResponseMode: "status",
		ParamsMode:   "json",
		Transport:    tr,
		Client:       client,
	}
```

mnezerka:
Hi all, I see issue is closed, but no there is no fix in codebase. I see same behavior as @wattanakorn495. Mosquitto consumes all sockets (~54K) and is not able to open any HTTP requests towards authentication HTTP service. What was the reason for closing this issue, please?

iegomez:
It was closed by the OP because he found a solution. As I wasn't able to reproduce the problem, I didn't make any changes to the code. I'm opening the issue again in case you can provide any more details on the error.

mnezerka:
[...]

PR #42, opened 2020-01-06
Move instance of http client to backend instance

mnezerka:
Keep one single instance of http client and reuse
it for all outgoing http requests. This change was
initiated by observation of resources leek: nubmer
of "closed" tcp socket file descriptors were incremented
by each set of auth requests (user, supervisor, acl)

This PR is related to #22

### DIFF
--- a/backends/http.go
+++ b/backends/http.go
@@ -26,6 +26,7 @@ type HTTP struct {
 	VerifyPeer   bool
 	ParamsMode   string
 	ResponseMode string
+	Client       *h.Client
 }
 
 type HTTPResponse struct {
@@ -109,6 +110,17 @@ func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
 		return http, errors.Errorf("HTTP backend error: missing remote options%s.\n", missingOpts)
 	}
 
+	// create instance of transport and client to be reused for all
+	// http calls
+	http.Client = &h.Client{Timeout: 5 * time.Second}
+
+	if !http.VerifyPeer {
+		tr := &h.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		http.Client.Transport = tr
+	}
+
 	return http, nil
 }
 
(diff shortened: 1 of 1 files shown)


backends/http.go
```diff
@@ -37,12 +39,21 @@ func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
 
 	log.SetLevel(logLevel)
 
+	tr := &h.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
iegomez: Whether peer verification is to be skipped should depend on the verifyPeer option.

backends/http.go
```diff
@@ -158,11 +169,11 @@ func (o HTTP) CheckAcl(username, topic, clientid string, acc int32) bool {
 		"acc":      []string{strconv.Itoa(int(acc))},
 	}
 
-	return httpRequest(o.Host, o.AclUri, username, o.WithTLS, o.VerifyPeer, dataMap, o.Port, o.ParamsMode, o.ResponseMode, urlValues)
+	return o.httpRequest(o.Host, o.AclUri, username, o.WithTLS, o.VerifyPeer, dataMap, o.Port, o.ParamsMode, o.ResponseMode, urlValues)
 
 }
 
-func httpRequest(host, uri, username string, withTLS, verifyPeer bool, dataMap map[string]interface{}, port, paramsMode, responseMode string, urlValues map[string][]string) bool {
+func (o HTTP) httpRequest(host, uri, username string, withTLS, verifyPeer bool, dataMap map[string]interface{}, port, paramsMode, responseMode string, urlValues map[string][]string) bool {
```
iegomez: This function no longer needs params that are associated to the HTTP backend as it's now a method of it and it can access them.

mnezerka:
@iegomez Thank you for review. I tried to address all your points in amended commit dfc14cf.

iegomez:
@mnezerka Thank you for taking the time.

Issue #43, opened 2020-01-14
Add option to supply authSource for auth_opt_mongo_username 


coldfire84:
With the jpmens plugin I was able to define a connection string that specified the authSource, as-per https://docs.mongodb.com/manual/reference/connection-string/#components.

I don't think I'm able to do this with the existing plugin options, snipped from the readme:
```
auth_opt_mongo_host 
auth_opt_mongo_port 
auth_opt_mongo_dbname 
auth_opt_mongo_username 
auth_opt_mongo_password 
auth_opt_mongo_users 
auth_opt_mongo_acls 
```

These options assume that the `auth_opt_mongo_username` exists in the same db as the users we wish to authenticate via the plugin. The user that plugin will connect with resides in my `admin` database, but the users I want the plugin to authenticate are in the `users` database, this difference  results in the following log entries on my mongodb instance:

```
2020-01-14T21:38:13.153+0000 I  ACCESS   [conn228] Supported SASL mechanisms requested for unknown user '<admin>@users'
2020-01-14T21:38:13.154+0000 I  ACCESS   [conn228] SASL SCRAM-SHA-1 authentication failed for <admin> on users from client <ip address>:<port>; UserNotFound: Could not find user "<admin>" for db "users"
```

I tried defining username as `<admin>@admin` but the mongodb log entries persist, albeit with the database name appended to the username:
```
2020-01-14T21:43:27.163+0000 I  ACCESS   [conn518] Supported SASL mechanisms requested for unknown user '<admin>@admin@users'
2020-01-14T21:43:27.164+0000 I  ACCESS   [conn518] SASL SCRAM-SHA-1 authentication failed for <admin>@admin on users from client <ip>:<port>; UserNotFound: Could not find user "<admin>@admin" for db "users"
```

iegomez:
Since the PR is merged and it's been mentioned elsewhere, I'll close this issue. @coldfire84 could you open another issue/PR for the documentation bits? Thanks!
BTW, would you be interested in being a collaborator so you can help me checking issues and PRs? No pressure though, you probably know by now that this moves slowly.

PR #44, opened 2020-01-14
Pr for #43, add mongodb option to specify AuthSource database name

coldfire84:
As per #43, this PR adds AuthSource, default value `""`, and creates IFTE so that:
* if specified AuthSource option (DB name) is supplied, use this in opts.Auth
* else, AuthSource follows original behavior, using DBName in opts.Auth

Also updated Docker file:
* Go version
* Mosquitto version
* Change build process to pull source from github, and package files did not appear to be up-to-date

### DIFF
--- a/backends/mongo.go
+++ b/backends/mongo.go
@@ -23,6 +23,7 @@ type Mongo struct {
 	Username        string
 	Password        string
 	DBName          string
+	AuthSource		string
 	UsersCollection string
 	AclsCollection  string
 	Conn            *mongo.Client
@@ -50,6 +51,7 @@ func NewMongo(authOpts map[string]string, logLevel log.Level) (Mongo, error) {
 		Username:        "",
 		Password:        "",
 		DBName:          "mosquitto",
+		AuthSource:		 "",
 		UsersCollection: "users",
 		AclsCollection:  "acls",
 	}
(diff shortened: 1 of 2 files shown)


coldfire84:
> Have you tested the changes?

Yes, I'm running them on my non-prod environment now.

> Could you add any necessary changes at the available tests to ensure this works as intended?

I can have a look at this, haven't written golang tests previously so should be a learning experience ... ! Essentially the code is written in such a way that it will default back to using dbname if authsource option is not supplied. 

So there is a new test scenario, but existing test should pass without issue.

> I think changes to the Dockerfile are a scope creep: they're unrelated to the issue and should not be included in the same PR.

Fine - it's your repo :)

iegomez:
I was asking for tests because I switched jobs a couple of months ago and thus been too busy to pay much attention to the plugin, nevermind mongodb which I've never used but to add support for it here. So I'd be really grateful if you could throw some more tests to ensure everything works correctly.

About the changes of the Dockerfile, I do like and appreciate them, but would also prefer to maintain them as a separate branches and PRs so any issue with one doesn't affect the other.

We can discuss both this directly once we connect on Slack too.

coldfire84:
Slack workspace [link](https://join.slack.com/t/cb-net/shared_invite/enQtODc1ODgzNzkxNTM3LTYwZGZmNjAxZWZmYTU4ZDllOGM3OTMxMzI4NzRlZmUzZmQ4NDljZWZiOTIwNTYzYjJmZjVlYzhhYWFiNThlMDA).

Issue #35, opened 2019-11-01
Question - comptibility of PBKDF2 passwords between jpmens plugin and this


sseide:
We used the `jpmens/mosquitto-auth-plug` mosquitto plugin with PBKDF2. Trying to switch the auth plugin to this one we cannot authenticated users anymore as passwords do not match.

As booth plugins use PBKDF2 with the same string format i thought passwords generated via jpmens `np` Tool or via NodeJS module `manolodd/mosquitto-pbkdf2` would work in your plugin too. Your `commons/HashCompare` function extracts all parts correctly but generates a different password hash, therefore authentication failes...

Can you explain whats different between your and jpmens implementaion/usage of PBKDF2 and suggest how to create a working password hash with NodeJS?

Thanks in advance,
Stefan Seide

iegomez:
Hi, Stefan! 
I'm sorry for not responding earlier. I don't have an answer right now but I'll look into it during the week. It does seem weird as before writing this plugin I also used `jpemns'` one and I made sure that passwords were compatible on switching, so maybe something changed later and I didn't notice it. So thanks for reporting the issue and I'll look into it as soon as possible.

sseide:
Thanks for answering - and no problem about time.

One difference i see so far is the fixed keylength (24) booth other projects use to generate the key:
Your library uses differnet sizes depnding on algorithm used (but never 24)

jpmens "np" tool:

https://github.com/jpmens/mosquitto-auth-plug/blob/34c1ab00ce22f0e32faf3a2563019fb97e60e687/np.c#L39-L41

and same inside the nodejs module to create compatible hashes or compare them:

https://github.com/manolodd/mosquitto-pbkdf2/blob/847cbc747a07245eb8048c5ab674d4d7f3578899/mosquitto_pbkdf2.js#L23-L29

But even setting the keylen to 24 at your code and recompiling it does not generate a match on comparing hashes created with jpmens or nodejs and your plugin....
Maybe handling of the password strings resp. hashes is different? Does Go interpret them as UTF-8 or something else?

Thanks for looking into it.

iegomez:
[...]

PR #45, opened 2020-01-17
PR mosquitto-auth-plug compatible hash

coldfire84:
As per #35 this PR includes the following option:
```
auth_opt_<backend-name>_salt_encoding utf-8
```
This option changes the way salt is converted to bytes and will result in a successful hash comparison if raw salt was not used in the generation of password hashes. If not supplied, this will default to 'base64' which is the default behavior of the plugin.

In order to for me to be able to test this I have had to include Mongo AuthSource code change. As a result you may want to close PR #44, and accept that this will also address #43 .

I'm yet to write tests, we can discuss approach/ code, but thought this would be useful to review in advance of our conversation.


iegomez:
Hey, Chris, hope everything's going OK! Have you had any time to check on this?

coldfire84:
Hey @iegomez , I plan on starting working through this PR (in terms of conflicts/ changes and testing in the coming week).

coldfire84:
@iegomez what's the best channel for us to discuss testing approach?

backends/files.go
```diff
@@ -59,6 +60,12 @@ func NewFiles(authOpts map[string]string, logLevel log.Level) (Files, error) {
 		return files, errors.New("Files backend error: no password path given")
 	}
 
+	if saltEncoding, ok := authOpts["salt_encoding"]; ok {
+		files.SaltEncoding = saltEncoding
+	} else {
+		files.SaltEncoding = "base64"
+	}
+
```
iegomez, line 68: You could set `base64` as default when creating the `files` var and then just set it to the encoding if given.
We should also check that the encoding is valid before assigning it.

backends/mongo.go
```diff
@@ -70,10 +73,21 @@ func NewMongo(authOpts map[string]string, logLevel log.Level) (Mongo, error) {
 		m.Password = mongoPassword
 	}
 
+	if saltEncoding, ok := authOpts["mongo_salt_encoding"]; ok {
+		m.SaltEncoding = saltEncoding
+		log.Infof("Mongo set salt encoding to: %s", saltEncoding)
+	} else {
+		m.SaltEncoding = "base64"
+	}
+
```
iegomez, line 82: Same as above.

backends/mongo.go
```diff
@@ -92,11 +106,21 @@ func NewMongo(authOpts map[string]string, logLevel log.Level) (Mongo, error) {
...
+				Password:    m.Password,
+				PasswordSet: true,
+			}
+		} else {
+			opts.Auth = &options.Credential{
+				AuthSource:  m.DBName,
+				Username:    m.Username,
+				Password:    m.Password,
+				PasswordSet: true,
+			}
```
iegomez, line 123: Similarly, you could just predefine `opts.Auth` and then assign its `AuthSource` conditionally.

backends/mysql.go
```diff
@@ -93,6 +94,12 @@ func NewMysql(authOpts map[string]string, logLevel log.Level) (Mysql, error) {
 		missingOptions += " mysql_password"
 	}
 
+	if saltEncoding, ok := authOpts["mysql_salt_encoding"]; ok {
+		mysql.SaltEncoding = saltEncoding
+	} else {
+		mysql.SaltEncoding = "base64"
+	}
+
```
iegomez, line 102: Same.

backends/postgres.go
```diff
@@ -77,6 +78,12 @@ func NewPostgres(authOpts map[string]string, logLevel log.Level) (Postgres, erro
 		missingOptions += " pg_password"
 	}
 
+	if saltEncoding, ok := authOpts["pg_salt_encoding"]; ok {
+		postgres.SaltEncoding = saltEncoding
+	} else {
+		postgres.SaltEncoding = "base64"
+	}
+
```
iegomez, line 86: Same.

backends/redis.go
```diff
@@ -43,6 +44,12 @@ func NewRedis(authOpts map[string]string, logLevel log.Level) (Redis, error) {
 		redis.Password = redisPassword
 	}
 
+	if saltEncoding, ok := authOpts["redis_salt_encoding"]; ok {
+		redis.SaltEncoding = saltEncoding
+	} else {
+		redis.SaltEncoding = "base64"
+	}
+
```
iegomez, line 52: Same.

backends/sqlite.go
```diff
@@ -58,6 +59,12 @@ func NewSqlite(authOpts map[string]string, logLevel log.Level) (Sqlite, error) {
 		sqlite.AclQuery = aclQuery
 	}
 
+	if saltEncoding, ok := authOpts["sqlite_salt_encoding"]; ok {
+		sqlite.SaltEncoding = saltEncoding
+	} else {
+		sqlite.SaltEncoding = "base64"
+	}
+
```
iegomez, line 67: Same.

common/utils.go
```diff
@@ -115,29 +113,60 @@ func hashWithSalt(password string, salt []byte, iterations int, algorithm string
...
 	buffer.WriteString(strconv.Itoa(iterations))
 	buffer.WriteString("$")
-	buffer.WriteString(base64.StdEncoding.EncodeToString(salt))
+	// Re-encode salt, using encoding supplied in saltEncoding param
+	if saltEncoding == "utf-8" {
+		buffer.WriteString(string(salt))
+
+	} else {
+		buffer.WriteString(base64.StdEncoding.EncodeToString(salt))
+	}
```
iegomez, line 122: Here we could have a check for the different encodings, while exported valid constants (or a map) could be used to check at backends initialization too, e.g.:
```
// Declare the valid encodings for validation.
const (
    UTF8 = "utf-8"
    Base64 = "base64"
)

...

switch saltEncoding{
  case UTF8:
    ...
  case Base64:
    ...
  default:
    // Error out here.
}
```

common/utils.go
```diff
@@ -115,29 +113,60 @@ func hashWithSalt(password string, salt []byte, iterations int, algorithm string
...
+		log.Errorf("HashCompare, invalid PBKDF2 hash supplied.")
 		return false
 	}
-
-	// Get the iterations and the salt and use them to encode the password
-	// being compared.cre
-	iterations, _ := strconv.Atoi(hashSplit[2])
-	salt, _ := base64.StdEncoding.DecodeString(hashSplit[3])
+	// Define empty error
+	var err error
```
iegomez, line 142: No need to.

common/utils.go
```diff
@@ -115,29 +113,60 @@ func hashWithSalt(password string, salt []byte, iterations int, algorithm string
...
+	iterations, err := strconv.Atoi(hashSplit[2])
+	if err != nil {
+		log.Errorf("Error getting number of iterations from PBKDF2 hash.")
+		return false
+	}
+	// Convert salt to bytes, using encoding supplied in saltEncoding param
+	salt := []byte{}
+	if saltEncoding == "utf-8" {
+		salt = []byte(hashSplit[3])
+	} else {
```
iegomez, line 153: Again, switch on the valid encodings.

iegomez:
@coldfire84 We could keep the discussion here while it works and move to a more direct flow if needed. Thanks again!

common/utils.go
coldfire84: Agreed, this was a legacy of an earlier approach where err was required elsewhere. 
Removed in latest commit.

backends/files.go
coldfire84: Added to latest commit.

backends/mongo.go
coldfire84: Added to latest commit.

backends/mongo.go
coldfire84: Modified approach in latest commit.

backends/mysql.go
coldfire84: Added to latest commit.

backends/postgres.go
coldfire84: Added to latest commit.

backends/redis.go
coldfire84: Added to latest commit.

common/utils.go
coldfire84: Modified approach used in most recent commit.

common/utils.go
coldfire84: Modified approach used in most recent commit.

backends/files.go
```diff
@@ -59,6 +61,18 @@ func NewFiles(authOpts map[string]string, logLevel log.Level) (Files, error) {
...
+		if saltEncoding == "base64"  {
+			files.SaltEncoding = saltEncoding
+			log.Infof("files backend: set salt encoding to: %s", saltEncoding)
+		} else if saltEncoding == "utf-8" {
+			files.SaltEncoding = saltEncoding
+			log.Infof("files backend: set salt encoding to: %s", saltEncoding)
+		} else {
+			log.Errorf("files backend: invalid salt encoding specified: %s", saltEncoding)
+		}
+	}
```
iegomez, line 74: Since they are already defined in the `common` package, we could just use the constants to switch over the encoding, e.g.:
```
switch saltEncoding{
		case common.Base64, common.UTF8:
			files.SaltEncoding = saltEncoding
			log.Debugf("files backend: set salt encoding to: %s", saltEncoding)
		default:
			log.Errorf("files backend: invalid salt encoding specified: %s, will default to base64 instead", saltEncoding)
}
```
Notice also that I changed the logging to debug level. This should be used in every backend.

coldfire84: Thanks, this is much cleaner, will add to other backends.

Am I right in thinking that debug logging will show no output by default? I'd set this to Info to ensure user got visual feedback that salt encoding was configured correctly.

iegomez: It won't show unless you explicitly set the log level to debug as part of the the options.

common/utils.go
```diff
@@ -115,29 +119,69 @@ func hashWithSalt(password string, salt []byte, iterations int, algorithm string
...
 	return buffer.String()
 }
 
 // HashCompare verifies that passed password hashes to the same value as the
 // passed passwordHash.
 // Taken from brocaar's lora-app-server: https://github.com/brocaar/lora-app-server
-func HashCompare(password string, passwordHash string) bool {
-	// SPlit the hash string into its parts.
+func HashCompare(password string, passwordHash string, saltEncoding string) bool {
+	//log.Debugf("Supplied:  ", passwordHash)
```
iegomez, line 142:
```suggestion
```

iegomez:
Everything looks good! The only thing we are missing is documenting changes in the Readme, but I' m fine if you wanna merge now and then add it. Thanks, @coldfire84!

iegomez:
@coldfire84 I took some time today to take care of a couple of issues and merged a small PR and have a second one ready to merge, and the latter's probably gonna introduce some minor conflicts. I'll gladly take over to merge those conflicts and also document the changes based on the original issues that you opened, but of course would also welcome docs coming from you since you know a lot more about MongoBD than I do.
Let me know and we can discuss it tomorrow.

Cheers!

coldfire84:
Will get the docs done tonight, apologies for the delays, work has been _mental!_

iegomez:
No worries and sorry for the pressure, was just trying to avoid any conflicts with upcoming changes. But then again, they shouldn't be terrible, so take your time (I know I have many times, so there's really no rush).

iegomez:
@coldfire84 Actually, I'll merge it right now since the other PR needs some more checking tonight. You may add documentation on a follow up PR. Thanks again!

PR #54, opened 2020-02-24
Docker Local Source

intolerance:
This pull request adds support for building a Docker container using the local source code.

I had to move this out to the root directory because `docker build` command will not let you traverse outside of it's pwd (e.g. ../).

In addition to the Dockerfile, I also added a lot of detail to the `README` in the `/docker` directory.

I hope you will consider merging this PR -- if I need to update or clean up anything, please let me know.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,9 @@ FROM debian:stable-slim as builder
 
 #Set mosquitto and plugin versions.
 #Change them for your needs.
-ENV MOSQUITTO_VERSION=1.6.3
+ENV MOSQUITTO_VERSION=1.6.8
 ENV PLUGIN_VERSION=0.6.1
-ENV GO_VERSION=1.12.6
+ENV GO_VERSION=1.13.8
 
 WORKDIR /app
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,70 @@
+
+#Use debian:stable-slim as a builder and then copy everything.
+FROM debian:stable-slim as builder
+
+#Set mosquitto and plugin versions.
+#Change them for your needs.
+ENV MOSQUITTO_VERSION=1.6.3
```
iegomez: Could you bump Mosquitto and Go versions to latest?

Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+RUN cd mosquitto-${MOSQUITTO_VERSION} && make WITH_WEBSOCKETS=yes && make install && cd ..
+
+#Get Go.
+RUN wget https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && tar -C /usr/local -xzf go${GO_VERSION}.linux-amd64.tar.gz
+RUN export PATH=$PATH:/usr/local/go/bin && go version && rm go${GO_VERSION}.linux-amd64.tar.gz
+
+
+# #Get the plugin.
+# RUN wget https://github.com/iegomez/mosquitto-go-auth/archive/${PLUGIN_VERSION}.tar.gz \
+#     && ls -l \
```
iegomez: Could you delete the commented bits here and later on this file?

Dockerfile
intolerance: Done.

Dockerfile
intolerance: I removed this section. Was this all you were referring to?

PR #66, opened 2020-05-14
Update README.md

ivica3730k:
Documentation (tutorial) specifies to use "auth-plugin.so"  when specifying the auth_plugin in .conf file for the mosquitto, but compiled binary file produced by following the tutorial is named "go-auth.so"

Is it possible this was a mistake which originated from the jpmens/mosquitto-auth-plug, as that project uses the auth-plugin.so name for the binary file?

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ and it is loaded into Mosquitto auth with the ```auth_plugin``` option.
 Set path to plugin and include conf.d dir for further configuration:
 
 ```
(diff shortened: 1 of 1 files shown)


iegomez:
Yeah, the name isn't too important, but it make sense to have the default one there. Thanks!

ivica3730k:
It just makes sense as it might prevent someone raising issue in the future if they just follow the tutorial blindly.

Thanks for the pull :)

PR #69, opened 2020-05-21
Minor fix in README.md

demestav:
I believe it should be `MOSQ_` instead of `MOSQL_`.


iegomez:
Yep, it should, thanks for catching that!

PR #79, opened 2020-06-18
Argon2id hashing and verification functions

garethhcoleman:
Hello

In this PR you'll find the basic code for an argon2id implementation, I've tested it in a very basic way and it seems to work... so now the real work begins!!! Obviously this new feature needs to be implemented in a way that extends existing functionality rather than breaking it, which is where I need some help.

As you can probably tell! I'm not an expert at the git stuff either and so I failed to squash my commits or take your recent commits into account, you can do this on merging apparently. I'm looking forward to benefiting from your experience and as you said on another pr - the disciplines you learn after developing a project for a while. I assume this will be going into a new feature branch?

Additionally as you suggested it would be nice to round out the hashing feature set with a bcrypt implementation as well. I see several previous issues raised around the password hashing function so I would also like to add some documentation to give additional guidance for the existing functionality as well as the new features. Several issues have included reference code in various languages for implementing matching hashing functions that would be nice to collect together to aid potential users.

So would you like to implement these features by extending the existing functions to cover multiple mechanisms, or by creating separate functions for each hashing type?
Looking forward to collaborating with you.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -109,7 +109,7 @@ func hashWithSalt(password string, salt []byte, memory uint32, iterations int, p
 	//log.Debugf("Generated: ", encodedHash())
 	return encodedHash
 }
-// HashCompare verifies that passed password hashes to the same value as the	
+// HashCompare verifies that passed password hashes to the same value as the
 // passed passwordHash.
 func HashCompare(password string, passwordHash string, saltEncoding string) bool {
 	hashSplit := strings.Split(passwordHash, "$")
--- a/common/utils.go
+++ b/common/utils.go
@@ -109,7 +109,8 @@ func hashWithSalt(password string, salt []byte, memory uint32, iterations int, p
 	//log.Debugf("Generated: ", encodedHash())
 	return encodedHash
 }
-
+// HashCompare verifies that passed password hashes to the same value as the	
+// passed passwordHash.
 func HashCompare(password string, passwordHash string, saltEncoding string) bool {
 	hashSplit := strings.Split(passwordHash, "$")
 
--- a/common/utils.go
+++ b/common/utils.go
@@ -112,6 +112,7 @@ func hashWithSalt(password string, salt []byte, memory uint32, iterations int, p
 // HashCompare verifies that passed password hashes to the same value as the
 // passed passwordHash.
 func HashCompare(password string, passwordHash string, saltEncoding string) bool {
+	// Split the hash string into its parts.
 	hashSplit := strings.Split(passwordHash, "$")
 
 	// Check array is of expected length
--- a/common/utils.go
+++ b/common/utils.go
@@ -18,7 +18,7 @@ import (
 
 // Declare the valid encodings for validation.
 const (
-    UTF8   = "utf-8"
+    UTF8 = "utf-8"
     Base64 = "base64"
 )
 
--- a/common/utils.go
+++ b/common/utils.go
@@ -18,8 +18,8 @@ import (
 
 // Declare the valid encodings for validation.
 const (
-	UTF8   = "utf-8"
-	Base64 = "base64"
+    UTF8   = "utf-8"
+    Base64 = "base64"
 )
 
 // OpenDatabase opens the database and performs a ping to make sure the
--- a/common/utils.go
+++ b/common/utils.go
@@ -1,28 +1,25 @@
 package common
 
 import (
-	"bytes"
 	"crypto/rand"
-	"crypto/sha256"
-	"crypto/sha512"
+	"crypto/subtle"
 	"encoding/base64"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
 
 	"github.com/pkg/errors"
-	"golang.org/x/crypto/pbkdf2"
+	"golang.org/x/crypto/argon2"
 
 	"github.com/jmoiron/sqlx"
 )
 
 // Declare the valid encodings for validation.
 const (
-    UTF8 = "utf-8"
-    Base64 = "base64"
+	UTF8   = "utf-8"
+	Base64 = "base64"
 )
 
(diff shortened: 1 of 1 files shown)


iegomez:
Hey, @garethhcoleman, thanks again for starting this!

I can't quite make the time today, but tomorrow I'll be sure to comment on your questions and at least start planning the work we need to get done.

Cheers!

iegomez:
Ok, we probably want to create a `Authroizer` (yeah, better names very welcome) interface that can both hash and compare, so something like:
```
type Authorizer interface {
  Hash(...params) (string, error)
  Compare(password string, passwordHash string, saltEncoding string) bool
}
```
Since each hashing method has different needed params we'll have to figure out what's the best way of abstracting that to the same interface, but shouldn't be much of a problem.

So then you'd have types for each that implement those methods:
```
type Pbkdf2 struct {
...
}

func (a *Pbkdf2) Hash(...params) (string, error) {
...
}

func (a *Pbkdf2) Compare(password string, passwordHash string, saltEncoding string) bool {
...
}

type Argon2 struct {
...
}

func (a *Argon2) Hash(...params) (string, error) {
...
}

func (a *Argon2) Compare(password string, passwordHash string, saltEncoding string) bool {
...
}

// And so on...

```

And finally, at `go-auth.go` we add an `authorizer common.Authorizer` (great opportunity to refactor those `common` and `utils`names) to the plugin and initialize it at `AuthPluginInit` depending on user options, and later inject it into backends so they may use it for comparison. Similarly, we would add an option to the `pw` utility to choose the `Authorizer` to generate passwords, and both the plugin an `pw` would default to the same one when no option is passed. That's if we go with a general `authorizer`, but we could also make them backend specific so each backend may use their preferred way.

How does that sound?

garethhcoleman:
That sounds good - is this a way to 'overload' the definitions of the functions Hash and Compare so that they have different params for each type of hash?

I think we should go the distance and create per-backend options, so that you could, for example, use argon2 for jtws and bcrypt for files. And yes, if options aren't passed you get the same behaviour as before, pbkdf2 with same keylength etc.

Names are one of the really hard things. Its my turn to say I haven't got much time free for a couple of days so I'll ponder on names in the background.

I've got a couple of other random ideas - like the option to strip username prefixes from other backends in the same way they are stripped for jwt's. But I can save that for after this and just put in a small PR.

iegomez:
1. It sounds like overloading, but really there's no overloading in Go like there's in, say, Java, where a class can redefine a method.  Well, in Go there's no inheritance either, so that makes sense, but there's no function overloading either. But just like in Java, you may define an interface that says how an implementer should "behave", that is what methods it needs to implement to qualify as a member of the club that interface is defining. The difference in this case is that in Java a class needs to explicitly implement an interface, while in Go if a struct implements the methods of an interface, then it effectively is part of the club: it looks like a duck, swims a like a duck, and quacks like a duck, then it's a duck; no need to explicitly say you are a duck.

2. I totally agree, the flexibility that per-backend options give is a must have if this plugin is to be useful: the only reason not to write your own is because the existing plugin supports whatever option you're looking for.

3. Well, what can I say? Naming is difficult and when I started this I knew very little Go and so I brought the usual Java conventions when names where needed, but I'll wait and see what you come up with.

4. You got me intrigued with those random ideas, I'm interested to know in which other cases stripping/trimming is useful. But that said, fair enough, I could really use a couple of days of rest too!

garethhcoleman:
Regarding names, as I understand it the `Authorizer` isn't doing authorizing really just the hashing and comparing, I suppose `HashCompare` or `HashCheck` might fit, although it would be nice if the name wasn't also part of the name of the function. Some languages use -able for interface names although a quick google suggests Go uses -er as you propose. In some ways the compare method is the key one as hash is a part of that but `comparer` is a bit generic so we could call it `HashComparer`?

Without the hashing and comparing, the utils.go package has OpenDatabase and matching topics functions. There is almost always some left over bits and pieces that don't belong cleanly anywhere else and so naming these is almost impossible because by definition there are no clear groups! `Common` is pretty good, I might prefer `Shared` as it makes it absolutely clear that this folder is for helper functions used across the codebase. And likewise I actually think `Utils` isn't a bad choice as we're unlikely to find a word for both matching and database opening functions.

iegomez:
`HashComparer` is what I was thinking too, so cool! We'll probably get a better feeling for it once implemented and seeing its use, but sounds right for now.

Regarding database opening and topic matching, they aren't used externally, so we should have no problem moving them over to the `backends` package. That way we can have a clean and dedicated package for hashing and comparing.

Thanks again for your insights and I hope to get real work done soon!

garethhcoleman:
Hello and excellent news if there is a natural place to absorb the left over functions in utile.go then we have deleted a whole folder! It is funny that as I get more experienced at software Dev I find sometimes more pleasure in deleting code than writing it! Even if I sweated over it a day or three ago, waking up to the realisation that it is unnecessary and I can just delete the redundant middle man method or whatever... Less is truly more!

I'll have a look at the bcrypt go library and canonical implementations over the next day or so and hopefully be able to have three central routines that we can then work with to fit inside a single called function. I hope you are happy with the idea of showing me how to bolt it into the environment variable parsing etc.

Thank you!

G

iegomez:
@garethhcoleman I'm starting to work on this right now, but since your branch removes the existing hash implementation I'll start a new branch and copy your changes to it. I'll then open a new PR and give you access in case.

iegomez:
I did the initial heavy lift work and opened a PR for it: https://github.com/iegomez/mosquitto-go-auth/pull/81/files

It is in no way complete, at the very least it needs initialization from the plugin, but also injecting it to the backends and testing each hasher. Nonetheless, I think the base is there and now we have a hashing package that can be extended and easily used throughout the plugin, so please review changes I did to your implementation of `argon2`, the new `bcrypt` implementation and anything else that may need a couple of fresh eyes.

PR #97, opened 2020-09-30
should be sslrootcert

AndreiD:
since you're setting up mysql_sslrootcert.....


backends/mysql.go
```diff
@@ -133,7 +133,7 @@ func NewMysql(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
 	}
 
 	if sslCert, ok := authOpts["mysql_sslrootcert"]; ok {
```
iegomez, line 135: Could you change `sslCert` here to also reflect this?

iegomez (commented):
You're absolutely right. :facepalm:

iegomez (approved):

Issue #100, opened 2020-10-21
Cache behavior for topics that are constantly active within the cache expiration time


arctic-alpaca:
Hello,

I'm using version 1.0.0 of this plugin and version 1.14.15 of Mosquitto in a docker container on a Raspberry Pi with redis as cache.
The plugin is used for authentication and authorization with the ChirpStack.

I noticed, when testing with very frequent published messages, that the cache doesn't expire even though cache expiration is set.
I'm not familiar with go, but I think this is caused by this function:
https://github.com/iegomez/mosquitto-go-auth/blob/4b20bb31764023df340e28dd4ab7303387b70401/cache/cache.go#L212

Every cache hit refreshes the expiration date. This leads to cache entries never getting removed even if the backend removes the entry. 
Is this intended? I think the cache should expire regardless of activity.

iegomez:
@arctic-alpaca Indeed, the cache is refreshed because it aims for the case of say having topics published with a frequency of ~15 seconds +- a couple of seconds, you can avoid having to hit your underlying store every couple of messages by setting the expiration to something slightly higher than that, ensuring quick availability for your frequent cases while not automatically granting for the less common ones.

But I see the need for a more general cache that just grants access for some period of time without refreshing, so thanks for bringing it up. Off the top of my head I'm thinking in options such as `no_refresh`, `max_refresh_time`, `max_refresh_amount`, etc. I'd love to know your opinion before implementing anything.

arctic-alpaca:
> Indeed, the cache is refreshed because it aims for the case of say having topics published with a frequency of ~15 seconds +- a couple of seconds, you can avoid having to hit your underlying store every couple of messages by setting the expiration to something slightly higher than that, ensuring quick availability for your frequent cases while not automatically granting for the less common ones.

While this does reduce load on the backend significantly, it also introduces a, in my opinion, quite problematic security issue. This behavior prevents an admin from revoking access to any user who publishes/receives messages with a frequency higher than the cache expiration. At least without restarting the service with `cache_reset true`.

I think the best approach would be to change the default behavior to let cache entries time out regardless of acivity and to introduce options to modify this behavior backend specific. For example the current behavior could be desired for ACL and password files, but is probably not desired for databases used to grant and revoke access dynamically.

iegomez:
[...]

PR #101, opened 2020-10-22
Cache refresh option

iegomez:
Adds `auth_opt_cache_refresh` option to control if the cache record expiration should be refreshed every time it's checked.
Fixes https://github.com/iegomez/mosquitto-go-auth/issues/100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,13 @@ There are 2 types of caches supported: an in memory one using [go-cache](https:/
 
 Set `cache` option to true to use a cache (defaults to false when missing) and `cache_type` to set the type of the cache. By default the plugin will use `go-cache` unless explicitly told to use Redis.
 Set `cache_reset` to flush the cache on mosquitto startup (**hydrating `go-cache` on startup is not yet supported**).
+
+
+**Update v1.2:**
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -248,12 +248,20 @@ There are 2 types of caches supported: an in memory one using [go-cache](https:/
 Set `cache` option to true to use a cache (defaults to false when missing) and `cache_type` to set the type of the cache. By default the plugin will use `go-cache` unless explicitly told to use Redis.
 Set `cache_reset` to flush the cache on mosquitto startup (**hydrating `go-cache` on startup is not yet supported**).
 
+
+**Update v1.2:**
```
arctic-alpaca, line 252: What do you think about adding a warning to ensure people are aware of what can happen if `cache_refresh true` is set? Maybe something like: 

Setting `cache_refresh true` in combination with continuously publishing/receiving messages more frequently than `auth_cache_seconds` and `acl_cache_seconds` can lead to cache entries getting "stuck" in the cache even though they aren't present in the backend any more.

iegomez: Thanks for the suggestion, I added a small warning about it.

PR #116, opened 2020-11-06
JWT checkers and JS mode

iegomez:
This PR adds:

- A JS interpreter runner, right now creating a VM on every check since it's not thread-safe. I'm probably gonna add a pool of VMs to improve it, but first I need to try it.
- A major cleanup of JWT backend, creating a `checker` interface that `js`, `local` and `remote` modes fulfill right now (not entirely ready). This makes it a lot easier to add the files-like check (https://github.com/iegomez/mosquitto-go-auth/pull/90), with or without the discussed custom syntax (bear with me guys, I'm getting there) as an option for this backend.
- A new Javascript backend.


iegomez:
@PierreF, this is the JWT changes I've working on. They are ready and tested, but I'd be happy to read your thoughts or any comments you may have before merging since they'll affect your work.

Once merged, my absolute priority is to review your PRs after adjusted.

Cheers!

iegomez:
And I just went along and added a separate JS backend of itself because it was straightforward.

PR #120, opened 2020-11-13
Improve handling of backend failure

PierreF:
This is more or less the port of https://github.com/jpmens/mosquitto-auth-plug/pull/296 and https://github.com/jpmens/mosquitto-auth-plug/pull/111

The issue this PR fix is is that when the HTTP (or MySQL) backend is down, ACL check return MOSQ_DENY_ACL. Which means that : 1) the message is rejected without notification for the sender 2) that rejection is cached.

This change few behavior:
* On failure, ACL check will reply to Mosquitto with MOSQ_ERR_UNKNOWN which will DISCONNECT the client (previous was deny the ACL). This is because denied ACL is silent, which could lead to data-loss
* HTTP (and JWT backends) will consider status code >= 500 as backend failure. An HTTP backend shouldn't use this to indicate a simple auth rejection.

Note on the PR changes:
* Most of the change is just returning the error and updating tests to verify that error is nil
* The C interface change, it return a uint8 instead of bool, since we need 3 states (granted, rejected, error). But this interface is only used between Go code and the auth-plugin.c from this repository I assume.
* I've made two unrelated change in jwt_test.go, both in "Given a token that is admin, super user should pass" for PostgreSQL and Mysql:
  * Both used `jwt.GetSuperuser(username)` (on line 198 and 360 on this PR), but I think `jwt.GetSuperuser(token)` is correct
  * Both changed `jwt.SuperuserUri = ""`, while I think `jwt.SuperuserQuery = ""` is correct.
  * Those change were not seen, because error (username isn't a JWT token) was converted to superuser = false


iegomez:
@PierreF As I mentioned in the other PR, I'll review next week and get back to you.

Cheers!

PierreF:
Hi,

is there something I can do to help this PR to get merged ?

iegomez:
@PierreF Sorry for taking so long, I've been prioritizing other things lately. As I mentioned in the README, I expect to get back to reviewing and working a bit on the plugin during this month and the next one.

iegomez:
@PierreF Hey again! I took care of a couple of bugs today and tried to get back to my JWT refactor, but in a very short span of time couldn't figure out exactly what I was doing a couple of months ago (yup, stupid me), just got to fix a couple of bugs in that branch and call it a day. I'm mentioning this because it is a major refactor in the JWT backend that could affect your PRs, so heads-up I guess.

Besides the anecdote, I'm really struggling to maintain the plugin lately and your PRs show both knowledge and will. I'm going on vacation in about a week, which may mean I finally get to review everything, or it may mean I don't have a proper connection and, once again, won't respond for a couple of weeks.

If it were the case I don't do anything for 3 weeks (and that's putting aside the month and a half of absence that already happened), is that critical to you? I gotta be honest: maintaining something that I haven't used for a really long time, is a burden. I've been busy with other things, and I'd be happy if you wanted to take over... or, if you want to, give a major help with the project. The JWT refactor is the only major thing I want to merge before being comfortable with the state of the plugin (with all its quirks and shit that need redoing, or, at the very least, rethinking) and being happy to give all the power to a peer to carry on.

Just shoot your thoughts and I'll gladly read them.

Cheers!

PierreF:
Thanks for keeping us informed.

At my company we need a Mosquitto authentication plugin. Since https://github.com/jpmens/mosquitto-auth-plug is archived (and I prefer writing Go than C :)), I'm willing to help maintaining this project.
While I think it's better to not be alone maintaining a software, I understand that without using a software its hard to find time for it (even when using it, it's not easy).

So to be clear, I'm open to both major help/being co-maintainer or taking alone the project.

Just tell me what you prefer and what are the next step.

iegomez:
Thanks, @PierreF! I'll keep you posted.

PierreF:
I'm fixing conflict generated by last merge. Should be pushed in the hour.

PierreF:
Conflict fixed. There is still one unrelated change in jwt_test.go: for test "But disabling superusers by removing superuri should now return false", the settings changed was "jwt_superquery" but I think the correct value is "jwt_pg_superquery" or "jwt_mysql_superquery".

iegomez:
You're absolutely right, missed that when changing general queries to DB specific ones. Nice catch!

backends/mysql.go
```diff
@@ -215,63 +215,63 @@ func NewMysql(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
...
 	err := o.DB.Get(&pwHash, o.UserQuery, username)
 
 	if err != nil {
 		log.Debugf("MySql get user error: %s", err)
-		return false
+		return false, err
 	}
 
 	if !pwHash.Valid {
 		log.Debugf("MySql get user error: user %s not found", username)
```
iegomez, line 229: Again, not on you, but this leaks. OTOH, this one is only in debug level, so maybe not that bad.

backends/mysql.go
```diff
@@ -215,63 +215,63 @@ func NewMysql(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
...
 	err := o.DB.Get(&count, o.SuperuserQuery, username)
 
 	if err != nil {
 		log.Debugf("MySql get superuser error: %s", err)
-		return false
+		return false, err
 	}
 
 	if !count.Valid {
 		log.Debugf("MySql get superuser error: user %s not found", username)
```
iegomez, line 258: Same story, I'll have to go through all these cases.

backends/sqlite.go
```diff
@@ -94,63 +94,63 @@ func NewSqlite(authOpts map[string]string, logLevel log.Level, hasher hashing.Ha
...
 	err := o.DB.Get(&pwHash, o.UserQuery, username)
 
 	if err != nil {
 		log.Debugf("SQlite get user error: %s", err)
-		return false
+		return false, err
 	}
 
 	if !pwHash.Valid {
 		log.Debugf("SQlite get user error: user %s not found.", username)
```
iegomez, line 108: Leak reminder.

iegomez (commented):
I left some comments for discussion, I still need to go through `go-auth.go` though.
I'll try to get to it today or tomorrow morning.

Also, I don't see any changes for the `gRPC` backend.

PierreF:
> Also, I don't see any changes for the `gRPC` backend.

Didn't you looked at a partial diff ? There are change to grpc.go file (and grpc_test.go).

iegomez:
> > Also, I don't see any changes for the `gRPC` backend.
> 
> Didn't you looked at a partial diff ? There are change to grpc.go file (and grpc_test.go).

Oops, sorry, I don't know how but I totally missed them.

backends/files.go
```diff
@@ -285,34 +285,34 @@ func checkCommentOrEmpty(line string) bool {
...
-		return false
+		return false, nil
 	}
 
 	if o.hasher.Compare(password, fileUser.Password) {
-		return true
+		return true, nil
 	}
 
 	log.Warnf("wrong password for user %s", username)
```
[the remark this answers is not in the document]
PierreF: I'm going to push another PR with those fix (or did you started any works on this ?) to avoid unrelated change in this (already big) PR.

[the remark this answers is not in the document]
iegomez: No problem, I'll open one later today.

backends/mongo.go
```diff
@@ -137,22 +137,22 @@ func (o Mongo) GetUser(username, password, clientid string) bool {
 	err := uc.FindOne(context.TODO(), bson.M{"username": username}).Decode(&user)
 	if err != nil {
 		log.Debugf("Mongo get user error: %s", err)
-		return false
+		return false, err
```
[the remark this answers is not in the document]
PierreF: Not yet read the driver documentation, but I'll add a test that show Mongo driver behave as you described: an error is returned if user isn't found.

go-auth.go
```diff
@@ -497,16 +519,31 @@ func setCache(authOpts map[string]string) {
...
+	}
+
+	return AuthRejected
+}
+
+func authUnpwdCheck(username, password, clientid string) (bool, error) {
 	var authenticated bool
 	var cached bool
 	var granted bool
+	var firstError error
```
iegomez, line 540: Sorry for bugging you with this, but could we rename it to `err`? There are 3 inner `err` appearances, but 2 are already scoped to their if from their construct, and the other one is being declared at place too (https://github.com/iegomez/mosquitto-go-auth/blob/f1a3fef9ee1aba22682ea5de46551188434431e7/go-auth.go#L601). Maybe that last one could be named `authErr` for clarity if you prefer.

go-auth.go
```diff
@@ -524,52 +561,77 @@ func AuthUnpwdCheck(username, password, clientid string) bool {
...
+	}
+
+	return AuthRejected
+}
+
+func authAclCheck(clientid, username, topic string, acc int) (bool, error) {
 	var aclCheck bool
 	var cached bool
 	var granted bool
+	var firstError error
```
iegomez, line 628: Same as above.

go-auth.go
```diff
@@ -661,8 +736,8 @@ func getPrefixForBackend(backend string) string {
 }
 
 //CheckBackendsAuth checks for all backends if a username is authenticated and sets the authenticated param.
-func CheckBackendsAuth(username, password, clientid string) bool {
-
+func CheckBackendsAuth(username, password, clientid string) (bool, error) {
+	var firstError error
```
iegomez, line 740: And here.

go-auth.go
```diff
@@ -675,20 +750,29 @@ func CheckBackendsAuth(username, password, clientid string) bool {
...
+
+	return authenticated, firstError
 
 }
 
 //CheckBackendsAcl  checks for all backends if a username is superuser or has acl rights and sets the aclCheck param.
-func CheckBackendsAcl(username, topic, clientid string, acc int) bool {
+func CheckBackendsAcl(username, topic, clientid string, acc int) (bool, error) {
 	//Check superusers first
+	var firstError error
```
iegomez, line 775: And here.

go-auth.go
```diff
@@ -712,36 +798,51 @@ func CheckBackendsAcl(username, topic, clientid string, acc int) bool {
...
 				aclCheck = true
 				break
+			} else if err != nil && firstError == nil {
+				firstError = err
 			}
 		}
 	}
 
-	return aclCheck
+	// If aclCheck is true, it means at least one backend didn't failed and
```
iegomez, line 811:
```suggestion
	// If aclCheck is true, it means at least one backend didn't fail and
```

go-auth.go
```diff
@@ -697,10 +781,12 @@ func CheckBackendsAcl(username, topic, clientid string, acc int) bool {
 			}
 			var backend = authPlugin.backends[bename]
 			log.Debugf("Superuser check with backend %s", backend.GetName())
-			if backend.GetSuperuser(username) {
+			if ok, err := backend.GetSuperuser(username); ok && err == nil {
 				log.Debugf("superuser %s acl authenticated with backend %s", username, backend.GetName())
 				aclCheck = true
 				break
+			} else if err != nil && firstError == nil {
```
iegomez, line 788: No need for `else if` if you're breaking in the previous clause.

go-auth.go
```diff
@@ -712,36 +798,51 @@ func CheckBackendsAcl(username, topic, clientid string, acc int) bool {
 			}
 			var backend = authPlugin.backends[bename]
 			log.Debugf("Acl check with backend %s", backend.GetName())
-			if backend.CheckAcl(username, topic, clientid, int32(acc)) {
+			if ok, err := backend.CheckAcl(username, topic, clientid, int32(acc)); ok && err == nil {
 				log.Debugf("user %s acl authenticated with backend %s", username, backend.GetName())
 				aclCheck = true
 				break
+			} else if err != nil && firstError == nil {
```
iegomez, line 805: No need for else if if you're breaking in the previous clause.

backends/mongo.go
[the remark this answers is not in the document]
iegomez: Looks like it's the case (from [the docs](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.3.3/mongo#Collection.FindOne)):

> If the filter does not match any documents, a SingleResult with an error set to ErrNoDocuments will be returned.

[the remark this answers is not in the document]
iegomez: So in this case we should check the error type to ensure we don't return an error when no document was found. Same goes for other DB backends that might return errors on unsuccessful queries.

iegomez (approved):
Thanks for dealing with non present users errors!
Remaining comments are only aesthetics related, but functionally this PR looks good to merge.

go-auth.go
PierreF: I need an `else if` because the `err` variable (now renamed to `getSuperuserErr`) is only defined in the if/else scope.

iegomez:
@PierreF feel free to merge when you're ready and I'll go on to review the other PRs.

PR #121, opened 2020-11-13
Add Docker image to run tests

PierreF:
To run tests (which requires lots of backends), I've used Docker containers.
This PR add steps to rebuild and run test using this Docker image.

It's probably not prefect (sometime Redis cluster fail... not every run) but it allow me to run the tests.

### DIFF
--- a/Dockerfile.runtest
+++ b/Dockerfile.runtest
@@ -1,0 +1,46 @@
+
+#Use debian:stable-slim as a builder and then copy everything.
+FROM debian:stable-slim as builder
+
+#Set mosquitto and plugin versions.
+#Change them for your needs.
+ENV MOSQUITTO_VERSION=1.6.10
+ENV PLUGIN_VERSION=0.6.1
+ENV GO_VERSION=1.13.8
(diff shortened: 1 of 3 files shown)


iegomez:
Hi, @PierreF! Thanks so much for this, I've been meaning to add something similar but never found the time. I won't be able to review it properly until next week since I won't have my dev machine during the weekend, but I'll take a look as soon as possible.

PierreF:
That's fine for me, thank for the query reply.

iegomez (commented):
Wait, isn't this missing `sqlite3` dependency? How is it passing? :thinking:

PierreF:
It's the same as base Dockerfile (just the end that change).

https://github.com/mattn/go-sqlite3 seems to bundle sqlite3 code in its repository (sqlite3-binding.c), which means sqlite3 get bundled in both source code and then in binary (static linking).

iegomez:
Oh, interesting! TIL.

iegomez (approved):

PR #123, opened 2020-11-13
Cache jitter

PierreF:
Port of https://github.com/jpmens/mosquitto-auth-plug/pull/315

Add a jitter to the expiration delay, which avoid lots of entry to expire at the same time. This could easily occur when the Mosquitto server is restarted: all clients get disconnected and will all reconnect quickly. This will cause a storm of lookups (can't avoid the first one, at least not without Redis). But without any jitter, every acl_cacheseconds a new lookup storm will happen.

This PR is based on #122 because otherwise a conflict will occur.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,12 +275,10 @@ auth_opt_auth_jitter_seconds 3
 auth_opt_acl_jitter_seconds 3
 ```
 
-The `auth_jitter_seconds` and `acl_jitter_seconds` allow to randomize the expiration used. The value used
-is cache_seconds +/- jitter_seconds. With above values (30 seconds for cache and 3 seconds for jitter), the
-expiration will be between 27 seconds and 33 seconds. The jitter is useful to reduce loopups storm that could
-occur every auth/acl_cache_seconds if lots of clients connected at the same time (for example
-after a server restart, all your clients may reconnect immediatly creating lots of entry expiring at the same time, unless
-a jitter is used). You can set jitter to 0 to disable this feature.
+`auth_jitter_seconds` and `acl_jitter_seconds` options allow to randomize cache expiration time by a given offset
+The value used for expiring a cache record would then be `cache_seconds` +/- `jitter_seconds`. With above values (30 seconds for cache and 3 seconds for jitter), effective expiration would yield any value between 27 and 33 seconds. 
+Setting a `jitter` value is useful to reduce lookups storms that could occur every `auth/acl_cache_seconds` if lots of clients connected at the same time, e.g. after a server restart when all clients may reconnect immediately creating lots of entries expiring at the same time.
+You may omit or set jitter options to 0 to disable this feature.
 
 If `cache_reset` is set to false or omitted, cache won't be flushed upon service start.
 
--- a/README.md
+++ b/README.md
@@ -271,7 +271,16 @@ auth_opt_cache_refresh true
 
 auth_opt_auth_cache_seconds 30
 auth_opt_acl_cache_seconds 30
+auth_opt_auth_jitter_seconds 3
+auth_opt_acl_jitter_seconds 3
 ```
+
+The `auth_jitter_seconds` and `acl_jitter_seconds` allow to randomize the expiration used. The value used
+is cache_seconds +/- jitter_seconds. With above values (30 seconds for cache and 3 seconds for jitter), the
+expiration will be between 27 seconds and 33 seconds. The jitter is useful to reduce loopups storm that could
+occur every auth/acl_cache_seconds if lots of clients connected at the same time (for example
+after a server restart, all your clients may reconnect immediatly creating lots of entry expiring at the same time, unless
+a jitter is used). You can set jitter to 0 to disable this feature.
 
 If `cache_reset` is set to false or omitted, cache won't be flushed upon service start.
 
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -6,6 +6,7 @@ import (
 	b64 "encoding/base64"
 	"fmt"
 	"hash"
+	"math/rand"
 	"strings"
 	"time"
 
@@ -19,6 +20,8 @@ import (
 type redisStore struct {
 	authExpiration    time.Duration
 	aclExpiration     time.Duration
+	authJitter        time.Duration
+	aclJitter         time.Duration
 	refreshExpiration bool
 	client            bes.RedisClient
 	h                 hash.Hash
@@ -27,6 +30,8 @@ type redisStore struct {
 type goStore struct {
 	authExpiration    time.Duration
 	aclExpiration     time.Duration
+	authJitter        time.Duration
+	aclJitter         time.Duration
 	refreshExpiration bool
 	client            *goCache.Cache
 	h                 hash.Hash
@@ -46,12 +51,14 @@ type Store interface {
 }
 
 // NewGoStore initializes a cache using go-cache as the store.
-func NewGoStore(authExpiration, aclExpiration time.Duration, refreshExpiration bool) *goStore {
+func NewGoStore(authExpiration, aclExpiration, authJitter, aclJitter time.Duration, refreshExpiration bool) *goStore {
 	// TODO: support hydrating the cache to retain previous values.
 
 	return &goStore{
 		authExpiration:    authExpiration,
 		aclExpiration:     aclExpiration,
+		authJitter:        authJitter,
+		aclJitter:         aclJitter,
 		refreshExpiration: refreshExpiration,
 		client:            goCache.New(time.Second*defaultExpiration, time.Second*(defaultExpiration*2)),
 		h:                 sha1.New(),
@@ -59,7 +66,7 @@ func NewGoStore(authExpiration, aclExpiration time.Duration, refreshExpiration b
 }
 
 // NewSingleRedisStore initializes a cache using a single Redis instance as the store.
-func NewSingleRedisStore(host, port, password string, db int, authExpiration, aclExpiration time.Duration, refreshExpiration bool) *redisStore {
+func NewSingleRedisStore(host, port, password string, db int, authExpiration, aclExpiration, authJitter, aclJitter time.Duration, refreshExpiration bool) *redisStore {
 	addr := fmt.Sprintf("%s:%s", host, port)
 	redisClient := goredis.NewClient(&goredis.Options{
 		Addr:     addr,
@@ -70,6 +77,8 @@ func NewSingleRedisStore(host, port, password string, db int, authExpiration, ac
 	return &redisStore{
 		authExpiration:    authExpiration,
 		aclExpiration:     aclExpiration,
+		authJitter:        authJitter,
+		aclJitter:         aclJitter,
 		refreshExpiration: refreshExpiration,
 		client:            bes.SingleRedisClient{redisClient},
 		h:                 sha1.New(),
@@ -77,7 +86,7 @@ func NewSingleRedisStore(host, port, password string, db int, authExpiration, ac
 }
 
 // NewSingleRedisStore initializes a cache using a Redis Cluster as the store.
-func NewRedisClusterStore(password string, addresses []string, authExpiration, aclExpiration time.Duration, refreshExpiration bool) *redisStore {
+func NewRedisClusterStore(password string, addresses []string, authExpiration, aclExpiration, authJitter, aclJitter time.Duration, refreshExpiration bool) *redisStore {
 	clusterClient := goredis.NewClusterClient(
 		&goredis.ClusterOptions{
 			Addrs:    addresses,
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -271,8 +271,17 @@ auth_opt_cache_refresh true
...
+auth_opt_acl_jitter_seconds 3
 ```
 
+The `auth_jitter_seconds` and `acl_jitter_seconds` allow to randomize the expiration used. The value used
+is cache_seconds +/- jitter_seconds. With above values (30 seconds for cache and 3 seconds for jitter), the
+expiration will be between 27 seconds and 33 seconds. The jitter is useful to reduce loopups storm that could
+occur every auth/acl_cache_seconds if lots of clients connected at the same time (for example
+after a server restart, all your clients may reconnect immediatly creating lots of entry expiring at the same time, unless
+a jitter is used). You can set jitter to 0 to disable this feature.
+
```
iegomez, line 284:
```suggestion
`auth_jitter_seconds` and `acl_jitter_seconds` options allow to randomize cache expiration time by a given offset
The value used for expiring a cache record would then be `cache_seconds` +/- `jitter_seconds`. With above values (30 seconds for cache and 3 seconds for jitter), effective expiration would yield any value between 27 and 33 seconds. 
Setting a `jitter` value is useful to reduce lookups storms that could occur every `auth/acl_cache_seconds` if lots of clients connected at the same time, e.g. after a server restart when all clients may reconnect immediately creating lots of entries expiring at the same time.
You may omit or set jitter options to 0 to disable this feature.

```
Besides the suggestion, this would only be true when refreshing the cache on startup, as a non refreshed cache wouldn't suffer from a horde of clients reconnecting, right? I think trusting the user to not set `jitter` when having a persistent cache is good enough, but do you think we should have some `no_jitter` default in that case?

cache/cache.go
```diff
@@ -115,6 +126,22 @@ func isMovedError(err error) bool {
 	return false
 }
 
+// return a expiration duration with a jitter added
+// The result is between expiration-jitter and expiration+jitter.
+// Negative value are never returned, 0 is used instead
```
iegomez, line 131:
```suggestion
// Return an expiration duration with a jitter added, i.e the actual expiration is in the range [expiration - jitter, expiration + jitter].
// If no expiration was set or jitter > expiration, then any negative value will yield 0 instead.
```
Curious if `jitter` > `expiration` should even be allowed, seems to me it should be strictly lower or equal. Thoughts?

cache/cache.go
```diff
@@ -239,29 +266,29 @@ func (s *redisStore) getAndRefresh(ctx context.Context, record string, expiratio
...
+	s.client.Set(record, granted, expirationWithJitter(s.authExpiration, s.authJitter))
 
 	return nil
 }
 
 //SetAclCache sets a mix, granted option and expiration time.
 func (s *goStore) SetACLRecord(ctx context.Context, username, topic, clientid string, acc int, granted string) error {
 	record := toACLRecord(username, topic, clientid, acc, s.h)
-	s.client.Set(record, granted, s.authExpiration)
+	s.client.Set(record, granted, expirationWithJitter(s.aclExpiration, s.aclJitter))
```
iegomez, line 277: Oopsies, nice catch!

cache/cache.go
```diff
@@ -239,29 +266,29 @@ func (s *redisStore) getAndRefresh(ctx context.Context, record string, expiratio
...
 	record := toAuthRecord(username, password, s.h)
-	return s.setRecord(ctx, record, granted, s.authExpiration)
+	return s.setRecord(ctx, record, granted, expirationWithJitter(s.authExpiration, s.authJitter))
 }
 
 //SetAclCache sets a mix, granted option and expiration time.
 func (s *redisStore) SetACLRecord(ctx context.Context, username, topic, clientid string, acc int, granted string) error {
 	record := toACLRecord(username, topic, clientid, acc, s.h)
-	return s.setRecord(ctx, record, granted, s.authExpiration)
+	return s.setRecord(ctx, record, granted, expirationWithJitter(s.aclExpiration, s.aclJitter))
```
iegomez, line 291: Again, nice catch! It does seem like you brought up these wrong expirations some time ago though, didn't you? If so, pulling `master` and rebasing would be needed.

cache/cache_test.go
```diff
@@ -8,17 +8,44 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestExpirationWithJitter(t *testing.T) {
+	// Since expirationWithJitter use random, to multiple time to ensure
+	// result is within expected boundary
+	for n := 0; n < 1000; n++ {
+		expiration := 100 * time.Millisecond
+		jitter := 0 * time.Millisecond
```
iegomez, line 16: Shouldn't we test with a non-zero jitter though? This would effectively use the expiration time only.

cache/cache_test.go
```diff
@@ -8,17 +8,44 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestExpirationWithJitter(t *testing.T) {
+	// Since expirationWithJitter use random, to multiple time to ensure
+	// result is within expected boundary
```
iegomez, line 13:
```suggestion
	/* Since expirationWithJitter randomizes the expirtaion time, and instructions take a random time already,
	we need to make room for tests.
	*/
```

cache/cache_test.go
```diff
@@ -8,17 +8,44 @@ import (
...
+		got = expirationWithJitter(expiration, jitter)
+		assert.True(t, 0 <= got)
+		assert.True(t, got <= expiration+jitter)
+	}
+}
+
 func TestGoStore(t *testing.T) {
 	authExpiration := 100 * time.Millisecond
 	aclExpiration := 100 * time.Millisecond
+	jitter := 0 * time.Millisecond
```
iegomez, line 38: Same as above. Also, could we have different `vars` for each `jitter`?

go-auth.go
```diff
@@ -374,6 +374,8 @@ func setCache(authOpts map[string]string) {
 
 	var aclCacheSeconds int64 = 30
 	var authCacheSeconds int64 = 30
+	var authJitterSeconds int64 = 3
```
iegomez, line 377: As almost always, this one's on me: `jitters` are only known to `cache`, there's no need to have them processed in the `main` package. Then again, `acl/authCacheSeconds` shouldn't be there either, but refactoring `cache` to process the option and extract the duration is totally out of scope for this PR, so leave it in `main` and we can take care of it later.

TL;DR: never mind this comment. And note to self: don't start open-source projects when you've only used the language for a couple of weeks, you'll regret it many years later when making/reviewing changes (and have a desire to burn the whole ugly, bad-designed thing down while we are at it). :woman_facepalming:

README.md
PierreF: This PR try to fix the following use case:
* You had hundred of clients connected
* You restart your mosquitto (like for upgrade).
* You don't have persistent cache
* All client got disconnected. All client reconnect (immediately in my case - at least within the few next seconds)
* All clients will trigger auth & acl checks at the same time. Since no persistent cache, it will storms the auth/acl backend. We can't avoid this one (without persistent cache).
* But without jitter, the cache expiration of all entry will be at the same time. Causing another storming of the backend on entries expiration (at least if client active enough, in my case the sent every 10 seconds).
* And this will continue every acl_cache_seconds

So:
* Yes this happend only if Mosquitto is restarted and mosquitto-go-auth don't have persistent cache
* (actually it happend also if for some reason lots of client connect at the same time - and stay connected)
* It don't try to fix the first storming, only subsequent
* With persistent cache, this jitter is likely not useful.

I'm not sure about the "`no_jitter` default" ? Do you just want that default value of jitter are 0 ? (which means no effect)

cache/cache.go
PierreF: Yes, it's likely not wanted and we should probably reject/warn for such "bad" configuration.

Because this function avoid returning negative number, it's not "broken" but it will result in odd behavior. At least the gain of a jitter add/subtract a random value it that the average expiration is `expiration`. This become wrong if `jitter` > `expiration`.

cache/cache.go
PierreF: hum... yes I did in PR #122 which is merged in master. But because this PR start from PR #122, I would have expected github to hide this diff once the first got merged.

cache/cache_test.go
PierreF: We do test with non-zero jitter just below. This one actually test that zero jitter disable the jitter behavior.

This test isn't needed in the loop (since it don't use randomization), do you want me to move it outside the loop ?

cache/cache_test.go
PierreF: I'm not sure to understand both your and my sentence :D
I likely did a typo and wanted to write "Since expirationWithJitter use random, DO multiple time [...]" or "Since expirationWithJitter use random, test multiple time [...]".

What I wanted to said it that since the result is randomized, we can't test for exact value but only that result is within a range. Doing multiple tests ensure that all results are within the range.

What about this version ?:
```
// Since expirationWithJitter randomizes the expirtaion time, do test multiple times and check that result is within expected range
```

cache/cache_test.go
PierreF: Well, I've added a `TestExpirationWithoutJitter` test which is a better place to test with `jitter = 0` :)

README.md
iegomez: What I meant was if e.g. the cache is persistent and not set to flush on restart, then we would default to no jitter even if the users provided one. But after some thought it may not be such a good idea.

cache/cache_test.go
iegomez: Ah, gotcha! Yeah, the `room` I mentioned was more about testing to the limit since we don't know how much the instructions take. I get what you were saying now and your new comment makes sense to me.

cache/cache.go
```diff
@@ -153,13 +179,13 @@ func (s *redisStore) Close() {
...
 	record := toAuthRecord(username, password, s.h)
-	return s.checkRecord(ctx, record, s.authExpiration)
+	return s.checkRecord(ctx, record, expirationWithJitter(s.authExpiration, s.authJitter))
 }
 
 //CheckAclCache checks if the username/topic/clientid/acc mix is present in the cache. Return if it's present and, if so, if it was granted privileges.
 func (s *goStore) CheckACLRecord(ctx context.Context, username, topic, clientid string, acc int) (bool, bool) {
 	record := toACLRecord(username, topic, clientid, acc, s.h)
-	return s.checkRecord(ctx, record, s.aclExpiration)
+	return s.checkRecord(ctx, record, expirationWithJitter(s.aclExpiration, s.authJitter))
```
iegomez, line 188:
```suggestion
	return s.checkRecord(ctx, record, expirationWithJitter(s.aclExpiration, s.aclJitter))
```

go-auth.go
```diff
@@ -384,6 +386,20 @@ func setCache(authOpts map[string]string) {
 		}
 	}
 
+	if authJitterSec, ok := authOpts["auth_jitter_seconds"]; ok {
+		authSec, err := strconv.ParseInt(authJitterSec, 10, 64)
+		if err == nil {
+			authJitterSeconds = authSec
+		} else {
+			log.Warningf("couldn't parse authJitterSeconds (err: %s), defaulting to %d", err, authJitterSeconds)
```
iegomez, line 394: This is kind of odd because I would assume that passing a wrong value would default to the same as omitting the option, i.e. 0, not 3. What do you think?

go-auth.go
```diff
@@ -393,6 +409,20 @@ func setCache(authOpts map[string]string) {
 		}
 	}
 
+	if aclJitterSec, ok := authOpts["acl_jitter_seconds"]; ok {
+		aclSec, err := strconv.ParseInt(aclJitterSec, 10, 64)
+		if err == nil {
+			aclJitterSeconds = aclSec
+		} else {
+			log.Warningf("couldn't parse aclJitterSeconds (err: %s), defaulting to %d", err, aclJitterSeconds)
```
iegomez, line 417: Same here.

iegomez (commented):
Almost done, just one fix and a discussion on default jitters on parsing error.

go-auth.go
PierreF: Either I'm missing something or I don't understand the first part of your comment.
For me (and my understanding of the code), the value used if option is omitted or in case of wrong value is 3.
Because in both case, we don't reach line 392 (`authJitterSeconds = authSec`) so we use using the default value set few line above.

I do agree that both omitted and wrong value must use the same default/fallback value.

I don't mind whether this default is 0 or 3. I've choose a default of jitter enabled because I think it only had benefit.

iegomez: Yeah, you're right that both are setting it to 3, I got thrown off by the docs which state that not passing a jitter would prevent the feature, i.e. the value should be 0, no 3. And if we're setting the default to 0, then I'd expect a parse error to default to 0 too.
Based on the principle of least surprise and if we wish to honor the docs, I think it makes sense to not have a jitter when not explicitly set by the user, and thus the default should be 0.

PierreF: Updated the default to 0.

iegomez (approved):

PR #124, opened 2020-11-13
Retry on backend error

PierreF:
Port of https://github.com/jpmens/mosquitto-auth-plug/pull/144

Allow to retry call to backend in case of error. This is useful if there are multiple backends behind a load-balancer and some backend are unhealthy (e.g. during canary deployment, where some backend use a new - bugged - version).

The default is to not retry, as this feature is not needed for every backend.

It's based on #120 because otherwise a conflict will occur. Since #120 contains lots of change, this PR should probably be only reviewed after #120 is merged (or only have a look at the 2nd commit).

This should be the last PR I made for this week :) Thanks for your works on this project, I hope all those PR won't give you too much work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,19 @@ auth_opt_log_file /var/log/mosquitto/mosquitto.log
 
 If `log_dest` or `log_file` are invalid, or if there's an error opening the file (e.g. no permissions), logging will default to `stderr`.
 
+#### Retry
+
+By default, if backend had an error (and no other backend granted access), an error is returned to Mosquitto.
+
+It's possible to enable retry, which will immediately retry all configured backends. This could be useful if the
+backend may be behind a load-balancer (like HTTP backend) and one instance may fail:
+
+```
+auth_opt_retry_count 2
+```
+
+The above example will do up to 2 retry (3 call in total) if backend had an error.
+
 #### Prefixes
 
 Though the plugin may have multiple backends enabled, there's a way to specify which backend must be used for a given user: prefixes. When enabled, `prefixes` allow to check if the username contains a predefined prefix in the form prefix_username and use the configured backend for that prefix. Options to enable and set prefixes are the following:
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ backend may be behind a load-balancer (like HTTP backend) and one instance may f
 auth_opt_retry_count 2
 ```
 
-The above example will do up to 2 retry (3 call in total) if backend had an error.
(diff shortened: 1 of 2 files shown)


PierreF:
This PR is based (or include) PR #120. I think once PR #120 got merged, the conflict will disappear. So doing nothing for now to avoid complicating git history.

README.md
```diff
@@ -365,6 +365,19 @@ auth_opt_log_file /var/log/mosquitto/mosquitto.log
...
+By default, if backend had an error (and no other backend granted access), an error is returned to Mosquitto.
+
+It's possible to enable retry, which will immediately retry all configured backends. This could be useful if the
+backend may be behind a load-balancer (like HTTP backend) and one instance may fail:
+
+```
+auth_opt_retry_count 2
+```
+
+The above example will do up to 2 retry (3 call in total) if backend had an error.
```
iegomez, line 379:
```suggestion
The above example will do up to 2 retries (3 calls in total considering the original one) if the responsible backend had an error or was down while performing the check.
```

auth-plugin.c
```diff
@@ -18,6 +18,11 @@
 
 #include "go-auth.h"
 
+// Same constant as one in go-auth.go.
+#define AuthRejected 0
```
iegomez, line 22: Wait, this seems odd, weren't these constants defined already in your previous PR?

auth-plugin.c
```diff
@@ -91,10 +96,16 @@ int mosquitto_auth_unpwd_check(void *userdata, const char *username, const char
   GoString go_password = {password, strlen(password)};
   GoString go_clientid = {clientid, strlen(clientid)};
 
-  if(AuthUnpwdCheck(go_username, go_password, go_clientid)){
+  GoUint8 ret = AuthUnpwdCheck(go_username, go_password, go_clientid);
+
+  if(ret == AuthGranted){
```
iegomez, line 101: Yup, confirmed, this should be a switch according to what we agreed in the previous PR. Mind rebasing the branch so those changes are taken into account before I go on reviewing?

iegomez (commented):
As mentioned, there are some discrepancies that should be solved with a rebase and force pushing the branch. I'll be able to properly review after that.

auth-plugin.c
PierreF: It's better once rebased on master. The original PR want forked from PR #120 and I hopped that Github would not longer show diff of #120 once merged.
It wasn't the case and clearly made this PR unreviewable.

iegomez (approved):

PR #136, opened 2020-12-16
example ACL values & working mysql query

simonfongnt:
added example ACL values as for permission.
updated working mysql ACL query for Testing example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,7 +386,7 @@ The main difference is that subscribe is checked at first, when a client connect
 So in practice you could deny general subscriptions such as # by returning false from the acl check when you receive `MOSQ_ACL_SUBSCRIBE`, but allow any particular one by returning true on `MOSQ_ACL_READ`.
 Please take this into consideration when designing your ACL records on every backend.
 
-Also, these are the current available values from `mosquitto`:
+Also, these are the current available bitwise values from `mosquitto`:
 
 ```
 #define MOSQ_ACL_NONE 0x00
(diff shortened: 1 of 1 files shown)


iegomez:
Hi, @simonfongnt, and thanks for your interest!

I wanna clarify a couple of things:
- The values are not _bitwise_ (isn't that and adjective for operators?), they are just `ints` that happen to be expressed in their `hex` form.
- A read/write permissions should not be `0x05`, **it might be** if it suits what you want from the query. In fact, there's no rule to how you build your queries, you just need to know what value Mosquitto will pass depending on the access request, e.g. `4` (or `0x04`) for `subscribe`. Then you may achieve whatever goal you want in ACL checks by either tweaking the query or the records.

simonfongnt:
@iegomez thanks for the feedback and I am sorry bothering you during your busy month. 

- I shall call it `binary flag` instead of `bitwise` but I am not convinced they are just happened to be in `hex` form.
- `0x05` should have read/subscription permission, `0x03` has read/write permission and `0x07` has all permissions and i had tested these.

if this is not the case, we may start over from issue and discuss it when you are available.

simonfongnt:
@iegomez after some digging in this and mosquitto projects, i can't find anything to support my binary flag statement. 
i suppose you are correct that it just happens to my query with `&` operator could have both read/subscription permission. 
i shall check again why my readonly ACL wasn't working before.
sorry to disturb you and please remove this pull request.

iegomez:
@simonfongnt No worries, you're not bothering at all! 
And as said, that may work, it's just not mandatory to use bitwise operations for auth.

Have a great end of year!

Issue #154, opened 2021-02-18
Files backend `topic` cannot contain <space character>.


mcr-ksh:
Hi,

there is a problem that MQTT standard allows whitespace characters, while the backend files.go does not support that due to the strings.Fields(line) splitting.

I used a workaround in my local copy:
`                                if len(lineArr) == 2 {
                                        aclRecord.Topic = strings.Replace(lineArr[1], "%20", " ", -1)
                                        aclRecord.Acc = MOSQ_ACL_READWRITE
                                } else {
                                        aclRecord.Topic = strings.Replace(lineArr[2], "%20", " ", -1)
`

I think this should be fixed properly.

Best,
CK

PierreF:
Hi,

thanks for the report. You solution also reply on using "%20" in the acl file instead of a real whitespace, isn't it ?

I've created a PR that fix it without needs to escape the whitespace (which is closer to what Mosquitto does, which is the format we said to support in our README). It however does not support multiple consecutive whitespaces in topic. Is it something you need ?

mcr-ksh:
Hi PierreF,

my solution is just a "hack" or workaround. Unfortunately we need multiple consecutive whitespaces.

Best,
CK

iegomez:
Hi, @mcr-ksh! Thanks for reporting this issue and sorry for not responding earlier, I've been having some issues with my personal machine and I don't like to use (nor I have setup correctly) my work one for personal development.

I see that @PierreF (sorry for not reviewing the other PRs yet, I have the same stated reason and having returned to work this week to blame for that) has already started working on this and I'll give it a look to see if I can give him a hand so we get this implemented ASAP: it surely is something that needs quick attention as it diverges from what Mosquitto's static files auth offers.

It's quite late here right now, so I'll try my best to tackle it tomorrow afternoon.

Cheers!

iegomez:
I just realized we also lack support for denying topics, i.e. `pattern deny <topic>`. I'll start to work on both concerns, but it'll obviously take some more time given the addition.

iegomez:
Ok, here's my take on the issue: https://github.com/iegomez/mosquitto-go-auth/pull/157
I still need to add tests to ensure it works as intended, but the logic to cover both cases is there. I'll add test cases as soon as I can.

iegomez:
@mcr-ksh @PierreF Merged and released: https://github.com/iegomez/mosquitto-go-auth/releases/tag/1.4.3

PR #155, opened 2021-02-21
Fix support for topic with whitespace

PierreF:
Fix for #154 with tests added and support for whitespace in topic and username.

It still do not work with topic (or username) that have multiple consecutive spaces (like user "my      username").

Supporting them make core more complex and/or require a 3rd party library (like strtok implementation to copy code logic that Mosquitto do). Since multiple consecutive spaces is topic or username seems a really edge case, I think we can kept a simpler code without 3rd party library and don't support this case.

### DIFF
--- a/backends/files.go
+++ b/backends/files.go
@@ -195,32 +195,29 @@ func (o *Files) readAcls() (int, error) {
 			continue
 		}
 
-		//If we see a user line, change the current user.
-		if strings.Contains(line, "user") {
-			//Try to get username
-			lineArr := strings.Fields(line)
+		lineArr := strings.Fields(line)
 
+		switch lineArr[0] {
+		//If we see a user line, change the current user.
+		case "user":
 			//Check format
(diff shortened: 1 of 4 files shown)


PierreF:
As @mcr-ksh said in #154, there is need for multiple consecutive whitespace. The solution in this PR does not works because Field() does not allow to known if there had one or more whitespace.

Using Split() (which allow to known the number of whitespace) isn't immediate, because line like `topic    read    test/topic/1` were supported by Field() but won't by (dummy) Split(). Becase the access mode is not longer the 2nd entry in the lineArr.

I'll try to implement a "SplitWordN()" function which split by word up to N word, such are `SplitWordN("topic     read    test/topic with space/1    ", 2) == ["topic", "read", "test/topic with space/1   "]`.
I'll check how Mosquitto handle leading/trailing whitespace in topic name to match it behavior.

PierreF:
Replaced by #157

PR #157, opened 2021-02-26
[Files]: Add support for deny ACL, and whitespaces in usernames/topics.

iegomez:


backends/files.go
```diff
@@ -349,20 +319,37 @@ func (o *Files) CheckAcl(username, topic, clientid string, acc int32) (bool, err
...
-			if TopicsMatch(aclRecord.Topic, topic) && (acc == int32(aclRecord.Acc) || int32(aclRecord.Acc) == MOSQ_ACL_READWRITE || (acc == MOSQ_ACL_SUBSCRIBE && topic != "#" && (int32(aclRecord.Acc) == MOSQ_ACL_READ || int32(aclRecord.Acc) == MOSQ_ACL_SUBSCRIBE))) {
-				return true, nil
+			match := TopicsMatch(aclRecord.Topic, topic)
+			if match {
+				if aclRecord.Acc == MOSQ_ACL_DENY {
+					return false, nil
+				}
+
+				if acc == int32(aclRecord.Acc) || int32(aclRecord.Acc) == MOSQ_ACL_READWRITE || (acc == MOSQ_ACL_SUBSCRIBE && topic != "#" && (int32(aclRecord.Acc) == MOSQ_ACL_READ || int32(aclRecord.Acc) == MOSQ_ACL_SUBSCRIBE)) {
+					return true, nil
```
PierreF, line 333: Shouldn't we process ann DENY before returning true ?

I think we should even process deny from both topic & pattern (`fileUser.AclRecords` and `o.AclRecords`) before any access, shouldn't we ?

backends/files.go
```diff
@@ -190,119 +204,75 @@ func (o *Files) readAcls() (int, error) {
...
 
-		//If we see a user line, change the current user.
-		if strings.Contains(line, "user") {
-			//Try to get username
-			lineArr := strings.Fields(line)
-
-			//Check format
-			if len(lineArr) == 2 && lineArr[0] == "user" {
-				_, ok := o.Users[lineArr[1]]
+		lineArr := strings.Fields(line)
```
PierreF, line 211: Due to Fields() we don't support username like `test       with       space`. Because after Fields() + Join() it will result in `test with space`.

Using `strings.Split()` would solve this issue but would introduce issue with line like `topic    read    topic/test/1`, becase `"read"` would not be `lineArr[1]`.

IMO, the two solution we can choose are:
* Don't support username / topic with multiple consecutive space. But from my understanding Mosquitto support this and this is requested (https://github.com/iegomez/mosquitto-go-auth/issues/154#issuecomment-783213816)
* Use smarter Split/Fields. I think writing some kind for `FieldsN()` that split/yield up to Nth fields with last being unsplitted is doable. We should then use FieldsN(line, 2) for user line and FieldsN(line, 3) for topic.

Ideally we should use a tokenizer (like Mosquitto does with strtok), from which we could ask the next fields or the rest of line. But such tokenizer seems more complex than writing FieldsN which seems enough (since we have only 2 kind of line: user or topic/pattern).

PierreF: The FieldsN I though is something like this: https://gist.github.com/PierreF/3866b140309b67e4507e4a52c691c353

backends/files.go
iegomez: Isn't this exactly that? https://github.com/iegomez/mosquitto-go-auth/pull/157/files#diff-e804824ad43abd98f9af0bed229f2641cc6f240856a1ff20d17e2599a4ba4e49R328

backends/files.go
iegomez: @PierreF Maybe it's github's formatting that's taking out some spaces, because both usernames look the same and the topic only shows one space between `topic` and `read`, and I believe you were showing that 2 spaces in between wouldn't rot, right?
I don't think is difficult to process it, but should we do that? The correct format is `topic read ...`, i.e. with one space, as far as I know.

iegomez: I may be wrong about that last statement though.

iegomez: Never mind, just checked Mosquitto's code and it indeed allows it.

backends/files.go
iegomez: Oooohhh, I just realized what you were saying and you were totally right. Sorry for that lapsus.

iegomez:
Ok, @PierreF , fixed the spaces and denials issues and added test for multiple spaces everywhere, at least I think I did. Also fixed another issue where topics for non existing users where being assigned to the last seen user instead of ignored (yikes, bugs everywhere!).

For spaces I went with a simple approach of pre-trimming the line and then removing the found prefix and re-trimming on next steps.

For example, for (assume underscores are spaces since they seemed to be formatted to a single one) `_topic__read___some_topic___with_any_number/of__spaces__`, the line would first be turned into `topic__read___some_topic___with_any_number/of__spaces`. After `topic` is read, then the line would be turned into `read___some_topic___with_any_number/of__spaces` and split on spaces to check the permission. After permission is checked, then the line gets the permission removed and trimmed again, i.e. `some_topic___with_any_number/of__spaces`, and that's saved as the topic for the ACL record. I could be missing something, but at least in my testing it works like a charm. Same goes for multiple spaces between `user` and the `username`.

iegomez:
And I just reviewed the other PRs. Yay!

backends/files.go
```diff
@@ -188,127 +203,123 @@ func (o *Files) readAcls() (int, error) {
...
+				permission := lineArr[0]
+
+				// Again, there may be more than one space between the permission and the topic, so we'll trim what's left after removing it and that'll be the topic.
+				line, err = removeAndTrim(permission, line, index)
+				if err != nil {
+					return 0, err
 				}
 
-				//Append to user or general depending on currentUser.
+				topic := strings.TrimSpace(line)
```
iegomez, line 273: Oh, this trimming looks redundant, I'll remove it later today.

backends/files.go
PierreF: I've updated my original comment. Didn't see github didn't allowed multiple space in `single quoted value`.

backends/files.go
```diff
@@ -188,127 +203,123 @@ func (o *Files) readAcls() (int, error) {
...
+				continue
 			}
-		} else if strings.Contains(line, "topic") {
 
-			//Split and check for read, write or empty (readwwrite) privileges.
+			userExists = true
+			currentUser = username
+		} else if prefix == "topic" || prefix == "pattern" {
+
 			lineArr := strings.Fields(line)
```
PierreF, line 236: This call is unneeded. We can reusing `lineArr` from outside the `if`, it has the same value (and it's reused after the `if`).

PierreF (approved):

PR #161, opened 2021-03-12
[Fix]: Build custom plugin on test

iegomez:


iegomez:
I bumped Go and Mosquitto versions to latest in the Docker test image. Couldn't get Mosquitto to compile correctly though, maybe manually building libwebsockets could fix it. :shrug: 
Since tests don't really in websockets, I preferred to build mosquitto without support in the meantime.

iegomez:
Oh, I also got a random redis-cluster failure when running test with docker, but haven't been able to reproduce and I've never seen it happen before. I'll try to look into it.

PR #170, opened 2021-04-28
Added monogdb TLS option

joda01:
@iegomez I needed TLS support for mongodb login. I added the option hard coded. Probably anyone else needes this option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1242,6 +1242,7 @@ auth_opt_mongo_users users_collection_name
 auth_opt_mongo_acls acls_collection_name
 auth_opt_mongo_disable_superuser true
 auth_opt_mongo_with_tls true
+auth_opt_mongo_insecure_skip_verify false
 ```
 
 The last two set names for the collections to be used for the given database.
@@ -1256,6 +1257,7 @@ When not set, these options default to:
 	users: 			 "users"
 	acls:  			 "acls"
     with_tls:        "false"
+    insecure_skip_verify: "false"
 
 If you experience any problem connecting to a replica set, please refer to [this issue](https://github.com/iegomez/mosquitto-go-auth/issues/32).
 
--- a/README.md
+++ b/README.md
@@ -1255,7 +1255,7 @@ When not set, these options default to:
 	dbame:           "mosquitto"
 	users: 			 "users"
 	acls:  			 "acls"
-    with_tls:             "false"
+    with_tls:        "false"
 
 If you experience any problem connecting to a replica set, please refer to [this issue](https://github.com/iegomez/mosquitto-go-auth/issues/32).
 
--- a/README.md
+++ b/README.md
@@ -1241,6 +1241,7 @@ auth_opt_mongo_password pwd
 auth_opt_mongo_users users_collection_name
 auth_opt_mongo_acls acls_collection_name
 auth_opt_mongo_disable_superuser true
+auth_opt_mongo_use_tls true
 ```
 
 The last two set names for the collections to be used for the given database.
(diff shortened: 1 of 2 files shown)


iegomez (changes requested):
Hi! I appreciate the contribution, but it needs a bit of work. It should guide the selection based on a new option to set TLS usage or not and others for any configuration needed, document said options, and also add tests.

joda01:
Okay, I will extend it.

joda01:
@iegomez Added config parameter. Is this correct implemented?

iegomez:
@joda01 Yeah, the option looks correctly set.

README.md
```diff
@@ -1254,6 +1255,7 @@ When not set, these options default to:
 	dbame:           "mosquitto"
 	users: 			 "users"
 	acls:  			 "acls"
+    tls:             "false"
```
iegomez, line 1258:
```suggestion
    with_tls:             "false"
```
Also, this looks not aligned.

backends/mongo.go
```diff
@@ -95,11 +98,25 @@ func NewMongo(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
...
+
+	if false == m.Tls {
+	  opts = options.ClientOptions {
+	    ConnectTimeout: &to,
+	  }
+	}else{
+	  opts = options.ClientOptions {
+	    ConnectTimeout: &to,
+	    TLSConfig:  &tls.Config{},
+	  }
```
iegomez, line 119:
```suggestion
	opts := options.ClientOptions {
	    ConectTimeout: &to,
	}
	
	if m.Tls {
	    opts.TLSConfig = &tls.Config{}
	}
```

backends/mongo.go
```diff
@@ -95,11 +98,25 @@ func NewMongo(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
...
+	var opts options.ClientOptions
+
+	if false == m.Tls {
+	  opts = options.ClientOptions {
+	    ConnectTimeout: &to,
+	  }
+	}else{
+	  opts = options.ClientOptions {
+	    ConnectTimeout: &to,
+	    TLSConfig:  &tls.Config{},
```
iegomez, line 118: I don't know which options this config might need for Mongo, could you check?

backends/mongo.go
```diff
@@ -29,6 +30,7 @@ type Mongo struct {
 	Conn             *mongo.Client
 	disableSuperuser bool
 	hasher           hashing.HashComparer
+	Tls              bool
```
iegomez, line 33:
```suggestion
	withTLS              bool
```
That would be consistent with how we call the option in `http` and `jwt` backends.

README.md
```diff
@@ -1241,6 +1241,7 @@ auth_opt_mongo_password pwd
 auth_opt_mongo_users users_collection_name
 auth_opt_mongo_acls acls_collection_name
 auth_opt_mongo_disable_superuser true
+auth_opt_mongo_use_tls true
```
iegomez, line 1244:
```suggestion
auth_opt_mongo_with_tls true
```
That would be consistent with how we call the option in `http` and `jwt` backends.

backends/mongo.go
joda01: It just needs this option `TLSConfig: &tls.Config{}`. It tells the connector to use TLS connection.
I have tested it with my mongodb and cosmosb mongo db driver.
You can find a full reference at https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo/options#ClientOptions

joda01 (commented):
@iegomez fixed your suggestions

iegomez (approved):
Looks good!

I'd just add the possibility to skip peer verification for testing purposes, see `InsecureSkipVerify` option at https://golang.org/pkg/crypto/tls/#Config. Of course, it should default to `false`, i.e. not skip it, but it's always a useful option for local testing with self signed certificates.

But it's not a blocker and could be added later, I'll leave it to you to decide.

And thanks again for the addition!

joda01:
Good suggestion! I added the skip_verify option. Can you please check again!

Thenks a lot!

PR #176, opened 2021-05-27
Add user-agent to requests

kfdm:
This avoids the default go `Go-http-client/1.1` User-Agent in requests.

Ideally I would have this as `mosquitto/<version>` though I'm unsure how to read the Mosquitto version yet. Any pointers?

### DIFF
--- a/backends/http.go
+++ b/backends/http.go
@@ -19,6 +19,7 @@ type HTTP struct {
 	UserUri      string
 	SuperuserUri string
 	AclUri       string
+	UserAgent    string
 	Host         string
 	Port         string
 	WithTLS      bool
@@ -77,6 +78,11 @@ func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
 	} else {
 		httpOk = false
 		missingOpts += " http_aclcheck_uri"
+	}
+
+	http.UserAgent = "mosquitto"
+	if userAgent, ok := authOpts["http_user_agent"]; ok {
+		http.UserAgent = userAgent
 	}
 
 	if host, ok := authOpts["http_host"]; ok {
@@ -224,6 +230,7 @@ func (o HTTP) httpRequest(uri, username string, dataMap map[string]interface{}, 
 		}
 
 		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("User-Agent", o.UserAgent)
 
 		resp, err = o.Client.Do(req)
 	}
--- a/backends/http.go
+++ b/backends/http.go
@@ -35,7 +35,7 @@ type HTTPResponse struct {
 	Error string `json:"error"`
 }
 
-func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
+func NewHTTP(authOpts map[string]string, logLevel log.Level, version string) (HTTP, error) {
 
 	log.SetLevel(logLevel)
 
@@ -80,7 +80,7 @@ func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
 		missingOpts += " http_aclcheck_uri"
 	}
 
-	http.UserAgent = "mosquitto"
+	http.UserAgent = fmt.Sprintf("%s-%s", defaultUserAgent, version)
 	if userAgent, ok := authOpts["http_user_agent"]; ok {
 		http.UserAgent = userAgent
 	}
--- a/backends/jwt_remote.go
+++ b/backends/jwt_remote.go
@@ -20,6 +20,7 @@ type remoteJWTChecker struct {
 	userUri      string
 	superuserUri string
 	aclUri       string
+	userAgent    string
 	host         string
 	port         string
 	withTLS      bool
@@ -78,6 +79,11 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions) (jwtC
 	} else {
 		remoteOk = false
 		missingOpts += " jwt_aclcheck_uri"
+	}
+
+	checker.userAgent = "mosquitto"
+	if userAgent, ok := authOpts["jwt_user_agent"]; ok {
+		checker.userAgent = userAgent
 	}
 
 	if hostname, ok := authOpts["jwt_host"]; ok {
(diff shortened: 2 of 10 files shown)


iegomez:
Hi, @kfdm!

Thanks for your contribution, but could you tell me why is this needed?

iegomez:
@kfdm We could probably pass the version on plugin initialization and then set it in http/jwt backend.

On the idea itself, I don't feel comfortable hardcoding the user agent. Instead, I'd add options that allow it to be set however the user wants, e.g.:
```
auth_opt_http_user_agent http-custom-user-agent
auth_opt_http_user_agent_mosquitto_version true 

auth_opt_jwt_user_agent jwt-custom-user-agent
auth_opt_jwt_user_agent_mosquitto_version true 
```

iegomez:
By the way, these are defined at `mosquito.h` header and provide the full version:
```
#define LIBMOSQUITTO_MAJOR 2
#define LIBMOSQUITTO_MINOR 0
#define LIBMOSQUITTO_REVISION 8
```

kfdm:
Thank you for the feedback! 🙇 

Though I have not implemented yet, instead of two fields how about a single value?

```
# No value to default to mosquitto
# custom value to set the agent
auth_opt_http_user_agent http-custom-user-agent
# Or a sentinel value to disable sending the agent all together
auth_opt_http_user_agent false
```

I am still trying to think of a clean way to get the version number. I feel like there is probably a cleaner way to handle the version than this, so I'm still thinking through it

```c
GoInt data[3] = {LIBMOSQUITTO_MAJOR, LIBMOSQUITTO_MINOR, LIBMOSQUITTO_REVISION};
GoSlice version = {data, 3, 3};

AuthPluginInit(keysSlice, valuesSlice, opts_count, version);
```

iegomez:
@kfdm I don't agree with defaulting to `mosquitto` as  the `User-Agent`, we shouldn't set one at all unless the user explicitly tells us so. When the option is not given, we simply shouldn't set the agent and let `go-http` do whatever it does, e.g. `Go-http-client/1.1` in the current version.

Likewise, since we're already adding options, I don't see a reason to impose the appended version: let the user decide if they want to attach it or not.

> I am still trying to think of a clean way to get the version number. I feel like there is probably a cleaner way to handle the version than this, so I'm still thinking through it

Mm... maybe a simple `sprintf` to concatenate and format the version as `"x.y.z"` and then pass it along as a `GoString`? Sorry, haven't given it much thought really. 🤷

iegomez:
@kfdm any updates on this? I like the general idea, but would like my last comment to be addressed. Let me know if you're available to do so, or if I should take care of it.

Cheers!

kfdm:
Thank you for checking back. I allowed myself to get distracted by some other things.

I feel like I have to disagree with you a bit on the default bit. I will yield to you on the specific version number, if you are worried about that from a security perspective, but as an operator, I feel like the least surprising behavior is for an application to provide a user agent. It can be rather frustrating on the infrastructure side, to try to track down a random script that uses that languages' default user-agent

iegomez:
@kfdm Yeah, you might be right. I don't really feel very strongly about it, maybe the fact that it'd be a change that would probably go unnoticed but could in fact surprise users (hey, why is there this user agent all of a sudden?) had me leaning towards not changing it by default. But let's just default to `mosquitto` and be explicit about it in the docs.

kfdm:
I've rebased to make sure there are no issues. There's not a specific changelog file in the repository itself, so I'm not sure if there's a specific place to call out the update to the default user-agent

backends/http.go
```diff
@@ -79,6 +80,11 @@ func NewHTTP(authOpts map[string]string, logLevel log.Level) (HTTP, error) {
 		missingOpts += " http_aclcheck_uri"
 	}
 
+	http.UserAgent = "mosquitto"
```
iegomez, line 83: Can we make the default a constant? Also, feel free not to export it, i.e. `userAgent` instead of `UserAgent`. I just haven't gotten to refactor `http` like I did with `jwt` backend.

backends/jwt_remote.go
```diff
@@ -80,6 +81,11 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions) (jwtC
 		missingOpts += " jwt_aclcheck_uri"
 	}
 
+	checker.userAgent = "mosquitto"
```
iegomez, line 84: Same here.

iegomez (commented):
Thanks for rebasing! Will you add Mosquitto's version? If so, please also add tests for both backends with and without having versions passed along.

iegomez:
> I've rebased to make sure there are no issues. There's not a specific changelog file in the repository itself, so I'm not sure if there's a specific place to call out the update to the default user-agent

I think adding to the README and documenting Mosquitto's version once you add it is enough, then I can mention the change ion the next release as I usually do.

kfdm:
I think I'll leave off the version number for now. Can always try to attempt it again, but my earlier attempts to try to get the version number from the mosquitto side were not as clean as I had hoped.

Both http.go and jwt_remote.go are in the same backend package, so I put `defaultUserAgent` with the rest of the constants. 

Would you prefer me to rebase everything into a single commit, or will you do that as part of the final merge?

iegomez:
Cool, `defaultUserAgent` looks good. And don't worry about rebasing, I'll give the version a try and add a test before squashing and merging.

Thanks again!

iegomez:
@kfdm I added the version to the default user agent in https://github.com/iegomez/mosquitto-go-auth/pull/182.
Let me know what you think.

kfdm:
That looks great! I'm not picky about getting credit for it. My work was quite simple. I'm also fine with you closing this PR and merging yours if that makes it easiest for you :)

iegomez:
It was your idea, @kfdm, I just added the version, so credit where credit is due.
I didn't know how to commit on your branch (is that even possible with PRs from forks?) so I opened a separate PR, but again, feel free to rebase and I'll merge yours. :wink:

kfdm:
Ok, I think I succeeded in rebasing correctly. It's now in two commits.
The first was my original few commits squashed into one
The second is your changes to bring in the version and all the test cases.

Git kept trying to assign the edit to me XD so I think I fixed that with `git commit --amend --author=".." --date="$(date -R)"`

Let me know if it still looks ok.

iegomez:
Cool then, let's merge it. Thanks, Paul!

iegomez (approved):

PR #177, opened 2021-05-27
Updated README

albertogeniola:
Added GO installation instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,8 +146,19 @@ WantedBy=multi-user.target
 If you are running another distro or need more details on building mosquitto, please check the offical mosquitto docs.
 
 #### Building the plugin
+Only Linux (tested in Debian, Ubuntu and Mint ùs) and MacOS are supported. 
 
-Only Linux (tested in Debian, Ubuntu and Mint versions) and MacOS are supported. This will build `go-auth.so` shared object:
+Before attempting to build the plugin, make sure you have go installed on the system. The required GO version for the current release is 1.13.8.
+
+```
+wget https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz
+sudo tar -C /usr/local -xzf go1.13.8.linux-amd64.tar.gz
+export PATH=$PATH:/usr/local/go/bin
+go version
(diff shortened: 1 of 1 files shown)


iegomez:
Hey, @albertogeniola! Thanks for your contribution.

I think if we want to bump up the versions we could use Mosquito 2.x.y. Also, the plugin doesn't depend on Go's 1.13 version, I'm pretty sure it runs fine with latest, though I haven't tried in a couple of weeks.

albertogeniola:
> Hey, @albertogeniola! Thanks for your contribution.
> 
> I think if we want to bump up the versions we could use Mosquito 2.x.y. Also, the plugin doesn't depend on Go's 1.13 version, I'm pretty sure it runs fine with latest, though I haven't tried in a couple of weeks.

Hi @iegomez ,
yeah, 1.13 is not mandatory. I've just picked that one as versions lower than that failed (due to missing TLS1.3). Which version would you advice to pin? I can test it and report back to you.
Also, I'm currently using mosquitto 2.0.9, so I agree we can bump that as well.

iegomez:
@albertogeniola maybe instead of pinning a version in instructions, you can just link to Go's download page: https://golang.org/dl/

albertogeniola:
In my opinion (and experience), we should at least specify or pin a minimum GO version. The reason why I started this PR is that: the installation instructions tells you to install go and so I did. However, after building problems, I realized that the go version was not recent enough: as soon as I updated go to 1.13, I was able to build. 

If you agree, I'd add some comments to the README to make users print the Go version and tell to update to minimum version 1.13 (via Go download website) if the installed version is older than that. Would this work for you?

albertogeniola (commented):
@iegomez : I've updated the instructions to point to the official Go download page. I've also bumped up the Go installed version to 1.16.

iegomez:
@albertogeniola My point was more about just linking to the official page instead of setting a specific version in the README, I absolutely agree about stating what's the minimum supported version.

iegomez (approved):

PR #178, opened 2021-06-02
Add 'tini' as init

kfdm:
A lot of containers try to avoid using `shell` for the ENTRYPOINT due to issues with escaping characters and not always handling signals very well. Many use [tini](https://github.com/krallin/tini#why-tini) as a way to handle this.

This is configured so that it is also easy to replace CMD with your own value without affecting the init.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ RUN export PATH=$PATH:/usr/local/go/bin && export CGO_CFLAGS="-I/usr/local/inclu
 FROM debian:stable-slim
 
 #Get mosquitto dependencies.
-RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid
+RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid tini
 
 #Setup mosquitto env.
 RUN mkdir -p /var/lib/mosquitto /var/log/mosquitto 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -38,7 +38,7 @@ RUN export PATH=$PATH:/usr/local/go/bin && export CGO_CFLAGS="-I/usr/local/inclu
 FROM debian:stable-slim
 
 #Get mosquitto dependencies.
-RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid
+RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid tini
```
iegomez, line 41:
```suggestion
RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid tiny
```

iegomez (changes requested):

iegomez:
@kfdm There's a typo that needs changing.
I don't know `tiny`, I'll look it up when I get some time before further reviewing.

Dockerfile
iegomez: Ha, sorry, the typo is the other way around.

docker/Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN mkdir mosquitto-go-auth && \
 FROM debian:stable-slim
 
 #Get mosquitto dependencies.
-RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid
+RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid tiny
```
iegomez, line 44:
```suggestion
RUN apt-get update && apt-get install -y libwebsockets8 libc-ares2 openssl uuid tini
```

iegomez (changes requested):

kfdm:
Woops, sorry about the typo there. I likely only tested the base Docker image and neglected to test the other one 🙇 
Should be fixed now.

This should explain things in detail https://github.com/krallin/tini/issues/8#issuecomment-146135930

iegomez (approved):

Issue #192, opened 2021-09-13
Mosquitto should start also without a GRPC connection


Vorlif:
Currently, Mosquitto cannot be started if the plugin cannot establish the GRPC connection. GRPC however, (can) uses a reconnect mechanism and establishes a connection independently. I realize that the plugin cannot be used if GRPC is configured and the remote site is unreachable. Withal, I don't think it should keep the entire service from launching just because the connection can't be established.

We had a problem that during a system start, Mosquitto was stopped again because the GRPC connection could not be established. But the GRPC service was active a few minutes later.

I am not sure if it is in line with the philosophy of the project, but I think it would be a useful change. Maybe as an additional option?

iegomez:
Hi, @Vorlif. Thanks for bringing this up, I think you're right and it makes sense to allow retries when attempting to create the client. I'll add it as an option.

iegomez:
I'll admit I haven't touched the gRPC backend in a couple of years now, so if you wanna contribute a PR adding the option instead, I'll happily review it.

Vorlif:
Hi @iegomez. Thank you for your quick reply. I would be happy to contribute something to the project. But it will take until the weekend before I can open a PR.

iegomez:
@Vorlif I had some time and came up with an initial attempt that allows to skip dial failure on initialization and only set the client on a user or all check one the service is up. I don't really have a gRPC service running to properly give it a try, but I added some basic automated tests to check it.

Could you take the branch for a spin to see if it works for you?

Vorlif:
@iegomez I have tested it and it works. :+1:  I left you also a suggestion as a comment in the PR.

Thanks for your effort and the project.

PR #193, opened 2021-09-14
Add gRPC options to set timeout and skip dial errors

iegomez:
- Fixes https://github.com/iegomez/mosquitto-go-auth/issues/192.
- Adds an option to set the dial timeout in ms, defaulting to 500 when not given or wrong (i.e., not an int).
- Adds an option to skip dial failures so they may be tried again on a user or all check, effectively allowing not to fail initialization but only auth checks until the service is up.
- Fixes - https://github.com/iegomez/mosquitto-go-auth/issues/194.

### DIFF
--- a/backends/grpc.go
+++ b/backends/grpc.go
@@ -57,8 +57,9 @@ func NewGRPC(authOpts map[string]string, logLevel log.Level) (*GRPC, error) {
 	tlsCert := []byte(authOpts["grpc_tls_cert"])
 	tlsKey := []byte(authOpts["grpc_tls_key"])
 	addr := fmt.Sprintf("%s:%s", authOpts["grpc_host"], authOpts["grpc_port"])
+	withBlock := authOpts["grpc_fail_on_dial_error"] == "true"
 
-	options, err := setup(addr, caCert, tlsCert, tlsKey)
+	options, err := setup(addr, caCert, tlsCert, tlsKey, withBlock)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


backends/grpc.go
```diff
@@ -165,13 +199,28 @@ func createClient(hostname string, caCert, tlsCert, tlsKey []byte) (*grpc.Client
 		})))
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
+	return nsOpts, nil
```
Vorlif, line 202: What do you think about removing the [`grpc.WithBlock()`](https://pkg.go.dev/google.golang.org/grpc#WithBlock) (optional) from the options? This would remove the need to also open the connection for the `grpc.DialContext` call. Furthermore, you could still set a backoff configuration and GRPC would take care of the connection itself. (By default, I believe, a backoff strategy is defined. )

So, you could run `grpc.DialContext` once and GRPC will take care of the connection on its own and reconnect if the connection was interrupted / not established.

Vorlif (approved):
LGTM

iegomez:
@Vorlif Thanks for taking a look. And yeah, that makes sense: as mentioned, I hadn't touched gRPC in a couple of years and just did a quick PoC to see if it worked, but totally missed the blocking dial option. I'll give the docs a better look to implement your suggestion.

Vorlif:
@iegomez It was a pleasure. I am also only superficially familiar with grpc, but have to do with it from time to time. So let me know if I can help you. 

Thanks for your work and your effort!

iegomez:
And now docker builds are working again.

Vorlif:
@iegomez Thank you for the adjustments. I have tested it and it works perfectly. I think the error logs are good, at least something is not working properly. Happy to see this in the master branch.

Vorlif (approved):
LGTM

PR #210, opened 2021-11-01
Add oauth2/openid-connect backend

tebartsch:
### Summary
The branch contains a new backend enabling authentication and authorization using an oauth2/openid-connect server.

Clients can either
 - hand over client credentials, i.e.  `client_id` and a `client_secret` ([client_credentials-oauth2-specification](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4)) or 
 - pass an `access_token` ([access_token-oauth2-specification](https://datatracker.ietf.org/doc/html/rfc6749#section-1.4))

**Note**
The code is based on [this](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2) project which implements the mosquitto-go-auth plugin backend.

### Access Token Authentication
If the backend receives the password `oauthbearer_empty_password` it assumes the username to be an `access_token`. Mosquitto seems to not allow empty passwords. If this can be changed the placeholder password could be removed. When using this option the the client has to reconnect using a new `access_token` after the old `access_token` has expired.

### Client Credentials Authentication
Any other combination of `username` and `password` is assumed to be a `client_id`/`client_secret`-pair and is used to obtain an `access_token`s from the authorization server regularly. Therefore client do not have to reconnect after token expiry in this case.

### Requesting Topic Permissions
In both cases the available `access_token` is used to request the servers `/userinfo`-endpoint ([userinfo-openid-specificaton](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo)). The backend expects the response of this request to look like
```
{
    "sub": "<username>",
    "mqtt": {
        "topics": {
            "read": ["/test/#", "/test/read/#],
            "write": ["/test/#","/test/write/#]
        },
        "superuser": false
    }
}
```
which contains the allowed topics for reading and writing corresponding to the provided `access_token`.

### Tests
Tests are implemented by using a mock oauth2 server.

### DIFF
--- a/backends/oauth2.go
+++ b/backends/oauth2.go
@@ -258,7 +258,7 @@ func (o *Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
 			log.Debug(err)
 			return nil, err
 		}
-		log.Debugf("Using Token: %s", token)
+		log.Debugf("Using Token: %s", token.AccessToken)
 	}
 	resp, err := client.Do(req)
 	if err != nil {
--- a/backends/oauth2.go
+++ b/backends/oauth2.go
@@ -251,7 +251,15 @@ func (o *Oauth2) createUserWithToken(accessToken, clientid string) (bool, error)
 
 func (o *Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
 	req, _ := http.NewRequest("GET", o.userInfoURL, nil)
-	log.Debugf("Performing request: %s %s:%s%s", req.Method, req.URL.Scheme, req.URL.Host, req.URL.Path)
+	log.Debugf("Performing request: %s %s://%s%s", req.Method, req.URL.Scheme, req.URL.Host, req.URL.Path)
+	if log.IsLevelEnabled(log.DebugLevel) {
+		token, err := client.Transport.(*go_oauth2.Transport).Source.Token()
+		if err != nil {
+			log.Debug(err)
+			return nil, err
+		}
+		log.Debugf("Using Token: %s", token)
+	}
 	resp, err := client.Do(req)
 	if err != nil {
 		log.Debug(err)
@@ -272,8 +280,6 @@ func (o *Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	log.Debugf("hey")
 
 	return &info, nil
 }
--- a/backends/oauth2.go
+++ b/backends/oauth2.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	. "github.com/iegomez/mosquitto-go-auth/backends/constants"
 	"github.com/iegomez/mosquitto-go-auth/backends/topics"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -46,13 +47,13 @@ type Oauth2 struct {
 	userCache               map[string]userState
 	cacheDuration           time.Duration
 	version                 string
-	scopesSplitted          []string
+	scopesSplit             []string
 }
 
-func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
+func NewOauth2(authOpts map[string]string, logLevel log.Level) (*Oauth2, error) {
 	log.SetLevel(logLevel)
 
-	var oauth2 = Oauth2{}
+	var oauth2 = &Oauth2{}
 
 	placedOpts := ""
 	missingOpts := ""
@@ -68,7 +69,7 @@ func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
 	}
 
 	oauth2.oauth2Config = go_oauth2.Config{
-		Scopes:      oauth2.scopesSplitted,
+		Scopes:      oauth2.scopesSplit,
 		RedirectURL: "",
 		Endpoint: go_oauth2.Endpoint{
 			TokenURL: tokenUrl,
@@ -77,7 +78,6 @@ func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
 	}
 
 	if userInfoURL, ok := authOpts["oauth_userinfo_url"]; ok {
-		placedOpts += "oauth_userinfo_url=" + userInfoURL + "\n"
 		oauth2.userInfoURL = userInfoURL
 	} else {
 		oauth2Ok = false
@@ -98,10 +98,10 @@ func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
 
 	if scopes, ok := authOpts["oauth_scopes"]; ok {
 		placedOpts += "oauth_scopes=" + scopes + "\n"
-		oauth2.scopesSplitted = strings.Split(strings.Replace(scopes, " ", "", -1), ",")
+		oauth2.scopesSplit = strings.Split(strings.Replace(scopes, " ", "", -1), ",")
 	} else {
 		log.Infof("no scopes where specified, using scope `all`")
-		oauth2.scopesSplitted = []string{"all"}
+		oauth2.scopesSplit = []string{"all"}
 	}
 
 	oauth2.userCache = make(map[string]userState)
@@ -115,7 +115,7 @@ func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
 	return oauth2, nil
 }
 
-func (o Oauth2) GetUser(username, password, clientid string) (bool, error) {
+func (o *Oauth2) GetUser(username, password, clientid string) (bool, error) {
 	// Get token for the credentials and verify the user
 	log.Debugf("Checking user with oauth plugin.")
 
@@ -126,7 +126,7 @@ func (o Oauth2) GetUser(username, password, clientid string) (bool, error) {
 	}
 }
 
-func (o Oauth2) GetSuperuser(username string) (bool, error) {
+func (o *Oauth2) GetSuperuser(username string) (bool, error) {
 	// Function that checks if the user has admin privilies
 	log.Debugf("Checking if user %s is a superuser.", username)
 
@@ -162,7 +162,7 @@ func (o Oauth2) GetSuperuser(username string) (bool, error) {
 	return cache.superuser, nil
 }
 
-func (o Oauth2) CheckAcl(username, topic, clientid string, acc int32) (bool, error) {
+func (o *Oauth2) CheckAcl(username, topic, clientid string, acc int32) (bool, error) {
 	// Function that checks if the user has the right to access to an address
 	log.Debugf("Checking if user %s is allowed to access topic %s with access %d.", username, topic, acc)
 
@@ -198,15 +198,15 @@ func (o Oauth2) CheckAcl(username, topic, clientid string, acc int32) (bool, err
 	return res, nil
 }
 
-func (o Oauth2) GetName() string {
+func (o *Oauth2) GetName() string {
 	return "OAuth Plugin " + o.version
 }
 
-func (o Oauth2) Halt() {
+func (o *Oauth2) Halt() {
 	// Do whatever cleanup is needed.
 }
 
-func (o Oauth2) createUserWithCredentials(username, password, clientid string) (bool, error) {
+func (o *Oauth2) createUserWithCredentials(username, password, clientid string) (bool, error) {
 	o.clientcredentialsConfig = go_clientcredentials.Config{
 		ClientID:     username,
 		ClientSecret: password,
@@ -234,7 +234,7 @@ func (o Oauth2) createUserWithCredentials(username, password, clientid string) (
 	return true, err
 }
 
-func (o Oauth2) createUserWithToken(accessToken, clientid string) (bool, error) {
+func (o *Oauth2) createUserWithToken(accessToken, clientid string) (bool, error) {
 	token := &go_oauth2.Token{
 		AccessToken: accessToken,
 		TokenType:   "Bearer",
@@ -263,7 +263,7 @@ func (o Oauth2) createUserWithToken(accessToken, clientid string) (bool, error) 
 	return true, err
 }
 
-func (o Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
+func (o *Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
 	req, _ := http.NewRequest("GET", o.userInfoURL, nil)
 	resp, err := client.Do(req)
 	if err != nil {
@@ -281,25 +281,32 @@ func (o Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
 	return &info, nil
 }
 
-func (o Oauth2) checkAccessToTopic(topic string, acc int32, cache *userState, username string, clientid string) bool {
+func (o *Oauth2) checkAccessToTopic(topic string, acc int32, cache *userState, username string, clientid string) bool {
 	log.Debugf("Check for acl level %d", acc)
 
 	// check read access
-	if acc == 1 || acc == 4 {
+	if acc == MOSQ_ACL_READ {
+		res := o.isTopicInList(cache.readTopics, topic, username, clientid)
+		log.Debugf("ACL for read was %t", res)
+		return res
+	}
+
+	// check subscribe access
+	if acc == MOSQ_ACL_SUBSCRIBE {
 		res := o.isTopicInList(cache.readTopics, topic, username, clientid)
 		log.Debugf("ACL for read was %t", res)
 		return res
 	}
 
 	// check write
-	if acc == 2 {
+	if acc == MOSQ_ACL_WRITE {
 		res := o.isTopicInList(cache.writeTopics, topic, username, clientid)
 		log.Debugf("ACL for write was %t", res)
 		return res
 	}
 
 	// check for readwrite
-	if acc == 3 {
+	if acc == MOSQ_ACL_READWRITE {
 		res := o.isTopicInList(cache.readTopics, topic, username, clientid) && o.isTopicInList(cache.writeTopics, topic, username, clientid)
 		log.Debugf("ACL for readwrite was %t", res)
 		return res
@@ -307,7 +314,7 @@ func (o Oauth2) checkAccessToTopic(topic string, acc int32, cache *userState, us
 	return false
 }
 
-func (o Oauth2) cacheIsValid(cache *userState) bool {
+func (o *Oauth2) cacheIsValid(cache *userState) bool {
 	log.Debugf("Cache Expiary: %s", o.cacheDuration)
 	log.Debugf("Last Update: %s", cache.updatedAt)
 	log.Debugf("Difference to now: %s", time.Now().Sub(cache.updatedAt))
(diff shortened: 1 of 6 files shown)


iegomez:
Hi, @imperaton, and thanks for the contribution!

Just a heads-up: I don't have the time to go over it right now, so it might take me some days to review it.

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+		return true
+	}
+	return false
+}
+
+func (o Oauth2) isTopicInList(topicList []string, searchedTopic string, username string, clientid string) bool {
+	replacer := strings.NewReplacer("%u", username, "%c", clientid)
+
+	for _, topicFromList := range topicList {
+		if topics.Match(replacer.Replace(topicFromList), searchedTopic) {
```
iegomez, line 330: Huh, this made me realize that maybe replacing patterns could be done in more general way.

backends/oauth2_test.go
```diff
@@ -0,0 +1,303 @@
...
+		authHeader := r.Header.Get("Authorization")
+
+		if authHeader == "" || authHeader == "Bearer wrong_token" {
+			http.Error(w, "Fail", 404)
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+
+		if authHeader == "Bearer mock_access_token_normaluser" {
+			w.Write([]byte("{\"sub\":\"mock_user_id_0\",\"mqtt\":{\"superuser\":false,\"topics\":{\"read\":[\"/test/topic/read/#\",\"/test/topic/writeread/1\",\"/test/topic/pattern/username/%u\",\"/test/topic/pattern/clientid/%c\"],\"write\":[\"/test/topic/write/+/db\",\"/test/topic/writeread/1\"]}}}"))
```
iegomez, line 25: The line length is quite big here and in other places. Maybe these could be properly defined structs and marshalled before writing them to the `ResponseWriter`, improving readability as well as making it easier to modify them later.

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	return &info, nil
+}
+
+func (o Oauth2) checkAccessToTopic(topic string, acc int32, cache *userState, username string, clientid string) bool {
+	log.Debugf("Check for acl level %d", acc)
+
+	// check read access
+	if acc == 1 || acc == 4 {
```
iegomez, line 288: Is there a reason not to allow for proper subscribe and read access and use them interchangeably?

iegomez (changes requested):

imperaton:
Thank you for the review. I'll add some modifications.

In case of your remarks regarding the cache and the pointer I have some further questions, as written above.

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+	clientcredentialsConfig go_clientcredentials.Config
+	tokenUrl                string
+	userInfoURL             string
+	userCache               map[string]userState
+	cacheDuration           time.Duration
+	version                 string
+	scopesSplitted          []string
+}
+
+func NewOauth2(authOpts map[string]string, logLevel log.Level) (Oauth2, error) {
```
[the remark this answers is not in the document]
imperaton: Of course, I'm not exactly sure about the pointer receiver method. I.e. What is the pointer receiver method for the Files backend?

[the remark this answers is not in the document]
iegomez: Notice how Files backend returns a pointer to a `Files` struct: https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/files.go#L18.

later, its methods are defined on a pointer receiver, e.g. https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/files.go#L47.

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+		Superuser bool `json:"superuser"`
+	} `json:"mqtt"`
+}
+
+type Oauth2 struct {
+	oauth2Config            go_oauth2.Config
+	clientcredentialsConfig go_clientcredentials.Config
+	tokenUrl                string
+	userInfoURL             string
+	userCache               map[string]userState
```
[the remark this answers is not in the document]
imperaton: Actually I was not yet aware of the general cache.

If I'd remove the specialized cache the plugin could make more requests to the oauth2-server than necessary, as far as I can tell. This happens if one client wants to read or write from multiple topics because the plugin would send a request to the server for every topic individually.

In contrast the specialized cache makes a single request and stores access permissions for all topics. If the same clients requests access for another topic that cache will be used.

Is there a way to reproduce this behavior using the general cache?

backends/oauth2.go
imperaton: I split them up. But for now I cannot distinguish between those two because the server only responds with the pattern

```
{
    "sub": "<username>",
    "mqtt": {
        "topics": {
            "read": ["/test/#", "/test/read/#],
            "write": ["/test/#","/test/write/#]
        },
        "superuser": false
    }
}
```

I could expand the server response to distinguish between `MOSQ_ACL_SUBSCRIBE` and `MOSQ_ACL_WRITE` (or even between all `MOSQ_ACL_NONE`, `MOSQ_ACL_READ`, `MOSQ_ACL_WRITE`, `MOSQ_ACL_READWRITE`, `MOSQ_ACL_SUBSCRIBE`, `MOSQ_ACL_DENY`). But for now this simple version is totally sufficient for me.

backends/oauth2.go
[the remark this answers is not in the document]
iegomez: Ah, indeed, you're caching the whole user information here, not just some read/written/subscribed topic.
I'll give it some thought and get back to you.

backends/oauth2.go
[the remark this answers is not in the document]
imperaton: Thank you. I've already done that by copying the structure from the files backend. See [commit](https://github.com/iegomez/mosquitto-go-auth/pull/210/commits/389a297c008172b316e08d05f6187c3a87af4700)

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+		oauth2.scopesSplitted = strings.Split(strings.Replace(scopes, " ", "", -1), ",")
+	} else {
+		log.Infof("no scopes where specified, using scope `all`")
+		oauth2.scopesSplitted = []string{"all"}
+	}
+
+	oauth2.userCache = make(map[string]userState)
+
+	if oauth2Ok {
+		log.Infof("OAuth Plugin initialized with configurations\n" + placedOpts)
```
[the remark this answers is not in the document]
imperaton: resolved in [commit](https://github.com/iegomez/mosquitto-go-auth/pull/210/commits/389a297c008172b316e08d05f6187c3a87af4700)

backends/oauth2.go
iegomez: Yeah, in order to merge this we'll need a general solution, it can be tied to a specific use case.

backends/oauth2_test.go
imperaton: i realized your recommendation in [commit](https://github.com/iegomez/mosquitto-go-auth/pull/210/commits/389a297c008172b316e08d05f6187c3a87af4700)

backends/oauth2.go
imperaton: Ok, then I would expect a server reponse of the form

```
{
    "sub": "<username>",
    "mqtt": {
        "topics": {
            "read": ["/test/#"],
            "write": ["/test/#"],
            "subscribe": ["/subscribe/#"],
            "deny": ["/deny0","/deny/#"],
        },
        "superuser": false
    }
}
```

Which should be enough to check for all access levels `MOSQ_ACL_NONE`, `MOSQ_ACL_READ`, `MOSQ_ACL_WRITE`, `MOSQ_ACL_READWRITE`, `MOSQ_ACL_SUBSCRIBE`, `MOSQ_ACL_DENY`.

backends/oauth2.go
```diff
@@ -0,0 +1,335 @@
...
+
+type Oauth2 struct {
+	oauth2Config            go_oauth2.Config
+	clientcredentialsConfig go_clientcredentials.Config
+	tokenUrl                string
+	userInfoURL             string
+	userCache               map[string]userState
+	cacheDuration           time.Duration
+	version                 string
+	scopesSplitted          []string
```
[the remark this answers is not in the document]
imperaton: resolved in [commit](https://github.com/iegomez/mosquitto-go-auth/pull/210/commits/389a297c008172b316e08d05f6187c3a87af4700)

backends/oauth2.go
imperaton: Implemented in [commit](https://github.com/iegomez/mosquitto-go-auth/pull/210/commits/b979ff9e4eccd1ad1250c3222ec33e1cc7fb3342).

Booyaabes:
Hi @imperaton,

I'm a user of [mosquitto-go-auth-oauth2](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2), and I'm really interested to have such a feature directly in `mosquitto-go-auth`.

I have some remarks on your PR:
- As scopes are optional, if no scope are provided, no scope should be send to Oauth2 server instead of 'all'. If user want 'all' scope, he should set `oauth_scopes` accordingly. I know that the same problem is in [mosquitto-go-auth-oauth2](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2) too,
- Your PR only support Client Credentials Grant type, which seems, in my opinion, really restrictive, as only confidential clients should use Client Credentials Grant type. I prefer the way the original plugin works, with a user context. Maybe you should allow both functions ?

tilmann-bartsch:
>     * As scopes are optional, if no scope are provided, no scope should be send to Oauth2 server instead of 'all'. If user want 'all' scope, he should set `oauth_scopes` accordingly. I know that the same problem is in [mosquitto-go-auth-oauth2](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2) too, I've just made a [PR](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2/pull/12),

Seems reasonable to me. I changed it.

>     * Your PR only support Client Credentials flow, which seems, in my opinion, really restrictive, as only confidential clients should use Client Credentials flow. I prefer the way the original plugin works, with a user context. Maybe you should allow both functions ?

In mosquitto-go-auth-oauth2](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2) two method are supported

- Using Client Credentials as mqtt-username and -password
- three-legged oauth2-flow using a oauth2-token as mqtt-username and an empty password 

see [GetUser-Function](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2/blob/8f07ab26f04e8f935f271885ef23048ac51a5361/src/main.go#L235-L244).

I replicated this behaviour with the single change, that the mqtt-password `oauthbearer_empty_password` must be specified if the mqtt-username should be interpreted as oauth2-token. I did this because Mosquitto doesnt seem to allow empty passwords anymore.

backends/oauth2.go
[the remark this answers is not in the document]
tilmann-bartsch: Hey @iegomez, did you reach a decision?

tilmann-bartsch:
I wonder if mosquitto would allow the go-auth-plugin to receive the contents of a mqtt 5.0 [AUTH](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901217)-Pacakage. This could be used by clients to send oauth2-tokens regularly so that the client does not have to reconnect if the current token expired.

Booyaabes:
> I replicated this behaviour with the single change, that the mqtt-password `oauthbearer_empty_password` must be specified if the mqtt-username should be interpreted as oauth2-token. I did this because Mosquitto doesnt seem to allow empty passwords anymore.

You're right, I missed those ones.

You can use Password grant type too: username + password + client id + client secret.
That's the flow I was talking about. So, it seems there is 3 methods supported.

backends/oauth2.go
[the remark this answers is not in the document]
iegomez: Sorry, I've been quite busy lately and haven't checked the plugin in a while. I'll try to go over the PR again this week.

Booyaabes:
> I just checked your code, it seems that it can handle password grant type. So it's only the documentation that does not contain all possible flows.
> 
> I'll try to test all the flows today or tomorrow.

My bad, password grant type doesn't work. For example, client secret read from config file is missing from your PR, see [here](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2/blob/master/src/main.go#L189).

tilmann-bartsch:
@Booyaabes Yes, I dropped the `username + password` flow in favor of the `client credentials` flow.

I did this because I needed a permanent method (when using token a reconnect is necessary periodically) for authentication/authorization of trusted services which access their own resources. The right method for this is the `client credentials see (i.e. see this [source](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)).

OAuth2 has been designed to not give an application hold on users `username/password`-combinations. This [source](https://www.oauth.com/oauth2-servers/access-tokens/password-grant/) states the method should not be used in practice.

If one one wants to tie mosquitto topic access to users (corresponding to persons) the application should obtain an access token by redirecting to keycloak ([authorization code grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1)). The use of such an access token is supported with the current state of the `mosquitto-go-auth` oauth2-backend.

Furthermore I cannot support the both methods at the same time, since `mosquitto-go-auth` can't tell if it received a client_id/client_secret- or username/password-pair. Therefore I am pleading for leaving the authorization flows as they are right now.

Booyaabes:
Yes, you're right @tilmann-bartsch.

tilmann-bartsch:
I tested the plugin again with a minimal setup to be sure everything works as aspected:

  https://github.com/tilmann-bartsch/mqtt-oauth2-test

@iegomez The Merge Request should be ready now.

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+func (o *Oauth2) GetUser(username, password, clientid string) (bool, error) {
+	if password == "oauthbearer_empty_password" {
+		return o.createUserWithToken(username, clientid)
+	} else {
+		return o.createUserWithClientCredentials(username, password, clientid)
+	}
+}
+
+func (o *Oauth2) GetSuperuser(username string) (bool, error) {
+	// Function that checks if the user has admin privilies
```
iegomez, line 115: The convention in Go is to annotate exported functions with a comment over the signature so the tooling may automatically convert them into documentation. Many times it's rather useless since the function/method name is clear enough, so we just do it to keep the formatter/linter from yelling. Having the comment inside the function is not really informative and should probably be removed.

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+		return false, err
+	}
+
+	log.Debugf("Check for superuser was %t", cache.superuser)
+	o.userCache[username] = cache
+	return cache.superuser, nil
+}
+
+func (o *Oauth2) CheckAcl(username, topic, clientid string, acc int32) (bool, error) {
+	// Function that checks if the user has the right to access a topic
```
iegomez, line 135: Same as above.

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+		return false, err
+	}
+
+	res := o.checkAccessToTopic(topic, acc, &cache, username, clientid)
+	log.Debugf("ACL check was %t", res)
+	return res, nil
+}
+
+func (o *Oauth2) GetName() string {
+	return "OAuth2 backend, version " + o.version
```
iegomez, line 155: I'm wondering if you're planning on adding support for more versions. Interesting. :smile:

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (o *Oauth2) Halt() {
+	// Do whatever cleanup is needed.
```
iegomez, line 159: I guess I'll trust you that there's nothing to do here. :smile:

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (o *Oauth2) updateCache(cache *userState) error {
+
+	if o.cacheIsValid(cache) {
```
iegomez, line 164: Mm... I'm not a fan of an `if/else` chain just to log something that could otherwise simply be logged in some kind of `cache valid: %s, o.cacheisValid(cache)` way.
Also, the following multiple calls to `log.Debugf` could just be formatted to be a single one with new lines in the format string.

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+	err := o.updateCache(&cache)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (o *Oauth2) getUserInfo(client *http.Client) (*UserInfo, error) {
+	req, _ := http.NewRequest("GET", o.userInfoURL, nil)
```
iegomez, line 254: We could use the exported `http` package request verbs instead of "GET".
Also, why are we bypassing the possible error? An http request can definitely fail!

backends/oauth2.go
```diff
@@ -0,0 +1,361 @@
...
+		return nil, err
+	}
+
+	return &info, nil
+}
+
+func (o *Oauth2) checkAccessToTopic(topic string, acc int32, cache *userState, username string, clientid string) bool {
+	log.Debugf("Check for acl level %d", acc)
+
+	if acc == MOSQ_ACL_NONE {
```
iegomez, line 291: Could we do a switch over the possible `acc` value instead of multiple `if` and `return`?

backends/oauth2_test.go
```diff
@@ -0,0 +1,338 @@
...
+	log "github.com/sirupsen/logrus"
+	. "github.com/smartystreets/goconvey/convey"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+	"time"
+)
+
+type oauth2ServerUserinfoResponse struct {
```
iegomez, line 16: Why do we need to declare a new struct that's almost the same us the one in the package being tested, just lacking one of the topics group?

backends/oauth2_test.go
```diff
@@ -0,0 +1,338 @@
...
+		for headerKey, headerValueList := range r.Header {
+			header += headerKey + "="
+			for _, value := range headerValueList {
+				header += value + ", "
+			}
+		}
+		log.Infof("Request Headers: %s", header)
+
+		accessToken := r.Form.Get("access_token")
+		// grantType := r.Form.Get("grant_type")
```
iegomez, line 94: Leftover comment here and probably in some other places.

iegomez (commented):
Hey, terribly sorry again for the time it takes me to take a look, I really haven't caught a breath lately.
I know I left a lot of comments and probably missed a bunch more since I was just skimming through the code when I got a minute, but I really appreciate the contribution and I think the functionality as a whole is there. They are mostly nitpicks and I'd fine with merging this PR and then having a follow up to tidy the backend up a bit.

What do yout think?

iegomez:
@tilmann-bartsch @Booyaabes sorry again for the time taken, I'm just checking on my latest review to see if you're still interested in going forward with this PR. If not, no worries and thank you again; I'll take a new look and try to finish it up in that case.

Cheers!

iegomez:
@tilmann-bartsch @Booyaabes I've been doing a cleanup of the repo these past few days and just wanted to check again on this.

tilmann-bartsch:
@iegomez Thank for the comments you made in one of the previous replies. Unfortunately I was not able to go forward with the PR in the last months and the situation won't change until September.

Still I would be happy to see the PR finished if you have the time for it.

PR #217, opened 2021-11-25
Bump mosquitto version 2.0.14

thomastvedt:
This PR only bumps mosquitto 2 version from 2.0.12 to 2.0.14
Changelog: https://mosquitto.org/ChangeLog.txt

Note:  
Perhaps it would be nice to automate this version bumping with Dependabot and get automatic pull requests when new versions are available? That would require some restructuring, since dependabot can't pick up the version in our `MOSQUITTO_VERSION_2 ` variable. But should be doable with some restructuring? 😊


iegomez (approved):

iegomez:
That would be quite nice indeed. For now, approved and merged.

PR #221, opened 2021-11-30
Add jwt remote option http method

fAuernigg:
* Add option to select http method.

### DIFF
--- a/backends/jwt_remote.go
+++ b/backends/jwt_remote.go
@@ -27,6 +27,7 @@ type remoteJWTChecker struct {
 	verifyPeer   bool
 
 	paramsMode   string
+	httpMethod   string
 	responseMode string
 
 	options tokenOptions
@@ -45,6 +46,7 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions, versi
 		verifyPeer:   false,
 		responseMode: "status",
 		paramsMode:   "json",
+		httpMethod:   "POST",
 		options:      options,
 	}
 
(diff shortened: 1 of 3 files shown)


backends/jwt_remote.go
```diff
@@ -63,6 +65,12 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions, versi
 		}
 	}
 
+	if httpMethod, ok := authOpts["jwt_http_method"]; ok {
+		if httpMethod == "POST" || httpMethod == "GET" || httpMethod == "PUT" {
```
iegomez, line 69: This is my bad as I never did before, but could I bother you to refactor the backend to use the `http` constants for methods?

iegomez (commented):
Sorry for the belated review.
The PR makes sense to me, but I'd like if we could introduce the same changes to the JWT remote one given their similarities.
Do you think you could add that?

backends/jwt_remote.go
iegomez: @fAuernigg I gather you lost interest in this PR: sorry, that's my bad for taking too long. Let me know if there's something you want to add or edit, elsewise I'll apply the same changes to the `http` backend (at the next point I get some time to do so) and merge.

Cheers!

fAuernigg: Sorry I missed to read your comment.
I will try to add the method to the http backend.

iegomez:
@fAuernigg I based on your PR and created https://github.com/iegomez/mosquitto-go-auth/pull/233 which just changes string literals to be taken from the http package constants, and adds a couple of tests. Hope you don't mind!

PR #223, opened 2021-12-14
Fix broken postgresql connection

thomastvedt:
See issue #218 

This PR adds max connection lifetime to postgresql db connection. Hopefully this will help with our issue where the auth plugin / mosquitto stops responding after a while (typically 30+ days..)

We're testing this version now in production, we'll test it for a while to verify that it helps on stability. => seems to be working 

https://github.com/iegomez/mosquitto-go-auth/issues/218
https://stackoverflow.com/questions/50338338/intermittent-connection-reset-by-peer-sql-postgres

I also bumped  
from: github.com/jmoiron/sqlx v1.2.0   
to: github.com/jmoiron/sqlx v1.3.4

PR #224, opened 2022-03-11
Postgresql fix

iegomez:
This PR completes work for https://github.com/iegomez/mosquitto-go-auth/pull/223.

@thomastvedt, feel free to check the additions and implement them in your fork if you wish; if so, I'll merge your PR instead.

Btw, I didn't add any tests because testing the life time of a connection as set by the `sqlx` package is out of my control and probably quite pointless, specially after you've confirmed it as a fix to your issue. That said, when adding the option to the three DB backends using said package, I left the default behaviour to not set any max life time as to not break backwards compatibility: that is, if you need to set a max life time, you need to be explicit in your options file, there'll be no default time set by the plugin (other than the `sqlx` set one, if any).


thomastvedt (approved):
Hi @iegomez , this looks great 👍 We can just close the other PR and merge this? 🙂

PR #230, opened 2022-05-08
Add option for sending decrypted claims to JWT Javascript backend

### LINKED ISSUE #229: Enhancement: Provide all JWT claims for backend
Why not send to `js`-type JWT backends all the claims, not just user name? It is not possible (or may be I just haven't found a way?) to use a third-party library in the JS script to parse JWT token string. In the same time if we need only the data from the JWT token to decide if we should allow user to access the topic, using a `remote`-type backend with the separate server seems to be an overkill. Providing the parsed data by the plugin itself looks like a reasonable solution.

ray66rus:
Closes #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -1488,6 +1488,7 @@ The `javascript` backend allows to run a JavaScript interpreter VM to conduct ch
 | js_user_script_path 		|				  |		Y		| Relative or absolute path to user check script		|
 | js_superuser_script_path	|				  |		Y		| Relative or absolute path to superuser check script	|
 | js_acl_script_path 		|				  |		Y		| Relative or absolute path to ACL check script			|
+| js_pass_claims	  	|    	false     |     N       | Pass all claims extracted from the token to check scripts		|
 
 This backend expects the user to define JS scripts that return a boolean result to the check in question.
 
@@ -1495,6 +1496,8 @@ The backend will pass `mosquitto` provided arguments along, that is:
 - `username`, `password` and `clientid` for `user` checks.
 - `username` for `superuser` checks.
 - `username`, `topic`, `clientid` and `acc` for `ACL` checks.
+If `js_pass_claims` option is set, an additional argument `claims` containing the claims data extracted
+from the JWT token is passed to all checks.
 
 
 This is a valid, albeit pretty useless, example script for ACL checks (see `test-files/jwt` dir for test scripts):
--- a/README.md
+++ b/README.md
@@ -880,21 +880,26 @@ The `jwt` backend is for auth with a JWT remote API, a local DB, a JavaScript VM
 
 The following `auth_opt_` options are supported by the `jwt` backend when remote is set to true:
 
-| Option            | default   | Mandatory | Meaning                            |
-| ----------------- | --------- | :-------: | ---------------------------------- |
-| jwt_host          |           |     Y     | API server host name or ip         |
-| jwt_port          |           |     Y     | TCP port number                    |
-| jwt_getuser_uri   |           |     Y     | URI for check username/password    |
-| jwt_superuser_uri |           |     N     | URI for check superuser            |
-| jwt_aclcheck_uri  |           |     Y     | URI for check acl                  |
-| jwt_with_tls      | false     |     N     | Use TLS on connect                 |
-| jwt_verify_peer   | false     |     N     | Whether to verify peer for tls     |
-| jwt_response_mode | status    |     N     | Response type (status, json, text) |
-| jwt_params_mode   | json      |     N     | Data type (json, form)             |
-| jwt_user_agent    | mosquitto |     N     | User agent for requests            |
-| jwt_http_method   | POST      |     N     | Http method used (POST, GET, PUT)  |
+| Option             | default   | Mandatory | Meaning                                                       |
+| ------------------ | --------- | :-------: | ------------------------------------------------------------- |
+| jwt_host           |           |    Y/N    | API server host name or ip                                    |
+| jwt_port           |           |     Y     | TCP port number                                               |
+| jwt_getuser_uri    |           |     Y     | URI for check username/password                               |
+| jwt_superuser_uri  |           |     N     | URI for check superuser                                       |
+| jwt_aclcheck_uri   |           |     Y     | URI for check acl                                             |
+| jwt_with_tls       | false     |     N     | Use TLS on connect                                            |
+| jwt_verify_peer    | false     |     N     | Whether to verify peer for tls                                |
+| jwt_response_mode  | status    |     N     | Response type (status, json, text)                            |
+| jwt_params_mode    | json      |     N     | Data type (json, form)                                        |
+| jwt_user_agent     | mosquitto |     N     | User agent for requests                                       |
+| jwt_http_method    | POST      |     N     | Http method used (POST, GET, PUT)                             |
+| jwt_host_whitelist |           |    Y/N    | List of hosts that are eligible to be an authoritative server |
 
 URIs (like jwt_getuser_uri) are expected to be in the form `/path`. For example, if jwt_with_tls is `false`, jwt_host is `localhost`, jwt_port `3000` and jwt_getuser_uri is `/user`, mosquitto will send a http request to `http://localhost:3000/user` to get a response to check against. How data is sent (either json encoded or as form values) and received (as a simple http status code, a json encoded response or plain text), is given by options jwt_response_mode and jwt_params_mode.
+
+if the option `jwt_parse_token` is set to `true`, `jwt_host` can be omitted and the host will be taken from the `Issuer` (`iss` field) claim of the JWT token. In this case the option `jwt_host_whitelist` is mandatory and must contain
+either a comma-separated list of the valid hostnames/ip addresses (with or without `:<port>` part) or the `*` (asterisk) symbol. If the `Issuer` claim is not contained in this list of valid hosts, the authorization will fail. Special
+value `*` means "any host" and is intended for testing/development purposes only - NEVER use this in production!
 
 If the option `jwt_superuser_uri` is not set then `superuser` checks are disabled for this mode.
 
--- a/backends/jwt_javascript.go
+++ b/backends/jwt_javascript.go
@@ -16,6 +16,8 @@ type jsJWTChecker struct {
 	superuserScript string
 	aclScript       string
 
+	passClaims bool
+
 	options tokenOptions
 
 	runner *js.Runner
@@ -79,6 +81,10 @@ func NewJsJWTChecker(authOpts map[string]string, options tokenOptions) (jwtCheck
 		return nil, errors.New("missing jwt_js_acl_script_path")
 	}
 
+	if passClaims, ok := authOpts["jwt_js_pass_claims"]; ok && passClaims == "true" {
+		checker.passClaims = true
+	}
+
 	checker.runner = js.NewRunner(checker.stackDepthLimit, checker.msMaxDuration)
 
 	return checker, nil
@@ -90,14 +96,10 @@ func (o *jsJWTChecker) GetUser(token string) (bool, error) {
 	}
 
 	if o.options.parseToken {
-		username, err := getUsernameForToken(o.options, token, o.options.skipUserExpiration)
-
-		if err != nil {
-			log.Printf("jwt get user error: %s", err)
+		var err error
+		if params, err = o.addDataFromJWT(params, token, o.options.skipUserExpiration); err != nil {
 			return false, err
 		}
-
-		params["username"] = username
 	}
 
 	granted, err := o.runner.RunScript(o.userScript, params)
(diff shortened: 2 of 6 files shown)


backends/jwt_javascript.go
```diff
@@ -90,14 +96,21 @@ func (o *jsJWTChecker) GetUser(token string) (bool, error) {
...
 		if err != nil {
-			log.Printf("jwt get user error: %s", err)
+			log.Printf("jwt get claims error: %s", err)
 			return false, err
 		}
 
-		params["username"] = username
+		if o.passClaims {
+			params["claims"] = claims
+		}
```
iegomez, line 108: [Nitpick] please add an empty line between the conditionals.

backends/jwt_javascript.go
```diff
@@ -90,14 +96,21 @@ func (o *jsJWTChecker) GetUser(token string) (bool, error) {
...
+		if username, found := claims[o.options.userFieldKey]; found {
+			params["username"] = username.(string)
+		} else {
+			params["username"] = ""
+		}
```
iegomez, line 113: This is used in the 3 checking methods, could probably be extracted to its own function/method.

conf_example/conf.d/go-auth.conf
```diff
@@ -1,14 +1,17 @@
...
-auth_opt_cache_db 4
+auth_opt_jwt_mode remote
+auth_opt_jwt_parse_token true
+auth_opt_jwt_secret qwertyuiopasdfghjklzxcvbnm123456
+
+auth_opt_jwt_params_mode json
+auth_opt_jwt_port 9000
+auth_opt_jwt_getuser_uri /check_user
+auth_opt_jwt_superuser_uri /check_superuser
+auth_opt_jwt_aclcheck_uri /check_acl
```
iegomez, line 17: Missing empty line at EOF.

conf_example/mosquitto.conf
```diff
@@ -3,14 +3,12 @@ persistence_location /var/lib/mosquitto/
...
 #keyfile /etc/letsencrypt/live/example.com/privkey.pem
 
-listener 1884
+listener 9001
 protocol websockets
 
 allow_anonymous false
-include_dir /etc/mosquitto/conf.d
\ No newline at end of file
+include_dir /mosquitto/conf/conf.d
```
iegomez, line 14: Missing empty line at EOF.

iegomez (changes requested):

iegomez:
Hey, @ray66rus!

Sorry for the belated review, I was a bit busy and didn't get any time to do so earlier.
The PR looks good overall, I just left a few nitpicks and would like to discuss passing the host in the issuer field.
What do you think about having a whitelist of hosts for it? In that case, my concern of raising privileges would be accounted for, at least I think so.

Thanks again for the contribution!

ray66rus:
> Sorry for the belated review, I was a bit busy and didn't get any time to do so earlier.

That's absolutely ok :)

> What do you think about having a whitelist of hosts for it? In that case, my concern of raising privileges would be accounted for, at least I think so.

Yes, I think it's a great idea. Unfortunately I missed this security concern. I'd also suggest to add a special value like "any" or "*" for testing/developing purposes with the appropriate warning against using it in production environment. So it will be a new option, like `jwt_host_whitelst` with possible values either `hostname1, hostname2, ..., hostname N` or `*`.

Where cd56b695e2a23b2a4d73d8c59dd33b9e7619f9b7 commit is concerned, I believe that it got to this MR by mistake - it was intended for the the in-house fork :) I can remove it altogether or leave it - if you think that these changes are relevant for other people.

iegomez:
Thanks, @ray66rus.

I like the idea of a wildcard when testing and having a warning logged when given. I would make sure to check for presence though, omitting the option or anything that's not `*` or a comma separated list of host, should error on initialization so people don't inadvertently start the plugin in an unsecured way.

About the Docker commit, there's another PR with some improvements, so let's remove it for now and we can revisit later after merging that PR.

README.md
```diff
@@ -1486,13 +1488,16 @@ The `javascript` backend allows to run a JavaScript interpreter VM to conduct ch
 | js_user_script_path 		|				  |		Y		| Relative or absolute path to user check script		|
 | js_superuser_script_path	|				  |		Y		| Relative or absolute path to superuser check script	|
 | js_acl_script_path 		|				  |		Y		| Relative or absolute path to ACL check script			|
+| js_pass_claims          |     false 	  |		  N   		| Pass all claims extracted from the token to check scripts		|
```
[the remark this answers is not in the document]
ray66rus: I've tried to fix it but note that here you use both spaces and tabs to format this table. So maybe a better way is to competely re-format this table with spaces only.

backends/jwt_javascript.go
ray66rus: Done

backends/jwt_javascript.go
ray66rus: Fixed

conf_example/mosquitto.conf
ray66rus: Irrelevant now - I removed this commit from the PR

ray66rus:
Well, I've fixed the issues you mentioned, removed a commit with changed `Dockerfile` and added whitelisting :)

README.md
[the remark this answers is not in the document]
iegomez: Oh, my bad then, sorry about that.

backends/jwt_remote.go
```diff
@@ -40,6 +42,10 @@ type Response struct {
 	Error string `json:"error"`
 }
 
+const (
+	whitelistMagicForAnyHost = "*"
```
iegomez, line 46: Haha, funny name. I'd call it `hostsWildcard` or something like that, but I kind of like it.

backends/jwt_remote.go
```diff
@@ -97,11 +103,36 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions, versi
...
+		checker.host = ""
 	} else {
 		remoteOk = false
 		missingOpts += " jwt_host"
 	}
 
+	if hostWhitelist, ok := authOpts["jwt_host_whitelist"]; ok {
+		if hostWhitelist == whitelistMagicForAnyHost {
+			log.Warning(
+				"Backend host whitelisting is turned off. This is not secure and should not be used in " +
```
iegomez, line 116: 👏

backends/jwt_remote.go
```diff
@@ -229,9 +260,31 @@ func (o *remoteJWTChecker) jwtRequest(host, uri, token string, dataMap map[strin
...
+		} else {
+			log.Printf("jwt remote get iss error: %s", err)
+		}
+	}
+
+	if host == "" {
+		errorString := fmt.Sprintf("cannot determine host for the authorization request for token %s", token)
+		log.Error(errorString)
+		return false, errors.New(errorString)
+	}
```
iegomez, line 283: You're double checking `host == ""`, so this could probably be refactored to something like:
```
if host == "" {
  if o.options.parseToken {
    // checking iss is in whitelist logic
  } else {
    // log error and bail out because we have no valid host, be it missing single one or not whitelisted
  }
}
```

backends/jwt_remote.go
```diff
@@ -229,9 +260,31 @@ func (o *remoteJWTChecker) jwtRequest(host, uri, token string, dataMap map[strin
...
+
+	fullURI := fmt.Sprintf("%s%s%s", tlsStr, host, uri)
 	if o.port != "" {
-		fullURI = fmt.Sprintf("%s%s:%s%s", tlsStr, o.host, o.port, uri)
+		fullURI = fmt.Sprintf("%s%s:%s%s", tlsStr, host, o.port, uri)
```
iegomez, line 287: What if the user passed a hosts whitelists, we matched one and it included a port, but mistakenly also set the `port` option?
We could have `getIssForToken` check for the presence of a port to set it in that case and avoid this problem. Thoughts?

README.md
```diff
@@ -880,22 +880,27 @@ The `jwt` backend is for auth with a JWT remote API, a local DB, a JavaScript VM
...
 
 URIs (like jwt_getuser_uri) are expected to be in the form `/path`. For example, if jwt_with_tls is `false`, jwt_host is `localhost`, jwt_port `3000` and jwt_getuser_uri is `/user`, mosquitto will send a http request to `http://localhost:3000/user` to get a response to check against. How data is sent (either json encoded or as form values) and received (as a simple http status code, a json encoded response or plain text), is given by options jwt_response_mode and jwt_params_mode.
 
+if the option `jwt_parse_token` is set to `true`, `jwt_host` can be omitted and the host will be taken from the `Issuer` (`iss` field) cliam of the JWT token. In this case the option `jwt_host_whitelist` is mandatory and must contain
+either a comman-separated list of the valid hostnames/ip addresses (with or without `:<port>` part) or the `*` (asterisk) symbol. If the `Issuer` claim contains any value not from this list, the authorization will failed. Special
+value `*` means "any name" and intended for the testing/development purposes only - NEVER use this value in the production environment.
```
iegomez, line 902:
```suggestion
if the option `jwt_parse_token` is set to `true`, `jwt_host` can be omitted and the host will be taken from the `Issuer` (`iss` field) claim of the JWT token. In this case the option `jwt_host_whitelist` is mandatory and must contain
either a comma-separated list of the valid hostnames/ip addresses (with or without `:<port>` part) or the `*` (asterisk) symbol. If the `Issuer` claim is not contained in this list of valid hosts, the authorization will fail. Special
value `*` means "any host" and is intended for testing/development purposes only - NEVER use this in production!
```

iegomez (commented):
Just a few more comments, but this is looking very good!

backends/jwt_remote.go
iegomez: This could come in handy to extract the hostname and the port parts: https://pkg.go.dev/net/url#URL

backends/jwt_remote.go
ray66rus: Yeah, valid point, thanks. I've refactored this piece of code into separate functions - IMHO, it looks much cleaner now.

BTW, note that even in the original version of your code you don't use "host" parameter in jwtRequest(). Maybe I'll remove it?

backends/jwt_remote.go
ray66rus: I've just added simple check for ":" in the "host" variable - it seems to be enough and more logical, since we are checking what we have in this variable in the same place where we are constructing URI for the request. But if you thnik that this is not a good enough solution, I can add something more sophisticated :)

iegomez: That seems fine for now, thanks.

backends/jwt_remote.go
```diff
@@ -323,3 +360,41 @@ func (o *remoteJWTChecker) jwtRequest(host, uri, token string, dataMap map[strin
...
+
+	return iss, nil
+}
+
+func (o *remoteJWTChecker) isHostWhitelisted(host string) bool {
+	if len(o.hostWhitelist) == 1 && o.hostWhitelist[0] == whitelistMagicForAnyHost {
+		return true
+	}
+
+	for _, whitelistedHost := range o.hostWhitelist {
```
iegomez, line 394: I just realized we could optimize this by having a `var hostsWhitelist map[string]struct{}` instead of a slice of them to get `O(1)` lookups, but I don't see the array being too big, so maybe not worth it for now.

backends/jwt_remote.go
iegomez: Could you point me to that part of the code to check?

iegomez:
I think this is pretty much ready, just want to check that host thing you mentioned before merging.

backends/jwt_remote.go
ray66rus: Well, in the origin "jwtRequest()" you use server host only in two places:
https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/jwt_remote.go#L232
https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/jwt_remote.go#L234

In both places you use a class member (`o.host`), not a function parameter (`host`). And you call this function always with the same class member as a parameter:
https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/jwt_remote.go#L157
https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/jwt_remote.go#L184
https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/jwt_remote.go#L212

iegomez: Ah, I see, my bad. Sorry about that. 🤦 
You're absolutely right that I was passing the `host` params from `o.host`, but then using `o.host` directly in the `jwtRequest` call, so it should be safe to remove it altogether from the arguments.

iegomez: So I could I bother you with that last bit to keep it all in this same PR and then we'll be ready to merge?
Sorry again.

iegomez: Nah, thank you for bearing with my late replies and doing all the work in this PR.

iegomez (approved):
:rocket: 
Haven't merged yet in case you wanna double check something or rerun tests locally, but if not, let me know and we'll merge.

iegomez:
Never mind, I deleted and refetched your branch to run tests and everything's ok, no idea why I couldn't simply pull latest changes, but anyway, it's working.. 
Which reminds me I need to set an action to run tests based on a contribution that allowed to execute them on a Docker image. I'm more familiar with Jenkins, a bit rusty on CircleCI, and new a bit of TravisCI on older times, but maybe some day I'll take the time to learn Github actions properly.

So I'm gonna merge now and prepare a minor release for it.

Thanks again, @ray66rus, this has been a wonderful contribution! If you have any more ideas or want to help with open issues (_cough_ ldap backend _cough_), I'll be happy to collaborate again and will try my best to be more responsive and helpful this time.

Cheers!

PR #234, opened 2022-06-08
Dockerfile upgrade

tetofonta:
Hi!

I've refactored the Dockerfile removing duplicate entries.
`libwebsocket` was compiled twice in two images but the second time was not required because the library is linked statically in the binary. (or at least it seems so, everything is working so...)
Also the final image now contains `mosquitto_*` utils and the required libraries (`libmosquitto.so` and `libmosquittopp.so`) enabling health checking of the container and some more scripting/automation during deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,13 +105,9 @@ COPY --from=mosquitto_builder /usr/local/bin/mosquitto_sub /usr/bin/mosquitto_su
 COPY --from=mosquitto_builder /usr/local/bin/mosquitto_pub /usr/bin/mosquitto_pub
 COPY --from=mosquitto_builder /usr/local/bin/mosquitto_rr /usr/bin/mosquitto_rr
 
-EXPOSE 1883 1884
-
-#RUN set -ex; echo "mosquitto_sub -t '\$SYS/#' -C 1 -p 1883 | grep -v Error || exit 1" > /healthcheck.sh
 RUN set -ex; ldconfig;
 
-#HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
-#    CMD ["/bin/sh", "-c", "/healthcheck.sh"]
+EXPOSE 1883 1884
 
 ENTRYPOINT ["/usr/bin/tini", "--"]
 CMD [ "/usr/sbin/mosquitto" ,"-c", "/etc/mosquitto/mosquitto.conf" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,9 @@ ARG MOSQUITTO_VERSION
 ARG LWS_VERSION
 
 # Get mosquitto build dependencies.
-RUN set -ex; apt-get update; apt-get install -y wget build-essential cmake libssl-dev libcjson-dev
+RUN set -ex; \
+    apt-get update; \
+    apt-get install -y wget build-essential cmake libssl-dev libcjson-dev
 
 # Get libwebsocket. Debian's libwebsockets is too old for Mosquitto version > 2.x so it gets built from source.
 RUN set -ex; \
@@ -35,13 +37,17 @@ RUN set -ex; \
 
 WORKDIR /app
 
-RUN set -ex; mkdir -p mosquitto/auth mosquitto/conf.d
+RUN mkdir -p mosquitto/auth mosquitto/conf.d
 
-RUN set -ex; wget http://mosquitto.org/files/source/mosquitto-${MOSQUITTO_VERSION}.tar.gz
-RUN set -ex; tar xzvf mosquitto-${MOSQUITTO_VERSION}.tar.gz
+RUN wget http://mosquitto.org/files/source/mosquitto-${MOSQUITTO_VERSION}.tar.gz
+
+RUN tar xzvf mosquitto-${MOSQUITTO_VERSION}.tar.gz
 
 # Build mosquitto.
-RUN set -ex; cd mosquitto-${MOSQUITTO_VERSION}; make CFLAGS="-Wall -O2 -I/build/lws/include" LDFLAGS="-L/build/lws/lib" WITH_WEBSOCKETS=yes; make install;
+RUN set -ex; \
+    cd mosquitto-${MOSQUITTO_VERSION}; \
+    make CFLAGS="-Wall -O2 -I/build/lws/include" LDFLAGS="-L/build/lws/lib" WITH_WEBSOCKETS=yes; \
+    make install;
 
 # Use golang:latest as a builder for the Mosquitto Go Auth plugin.
 FROM --platform=$BUILDPLATFORM golang:latest AS go_auth_builder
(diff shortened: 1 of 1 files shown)


iegomez:
Hey, @tetofonta, thanks for this!

I forgot I had to approve the CI run to check the solution, and there are also a couple of PRs I need to check upon before merging anything else, but I'll try to get back to you as soon as possible.

Dockerfile
```diff
@@ -122,7 +98,20 @@ COPY --from=go_auth_builder /app/pw /mosquitto/pw
...
+COPY --from=mosquitto_builder /usr/local/bin/mosquitto_rr /usr/bin/mosquitto_rr
+
 EXPOSE 1883 1884
 
+#RUN set -ex; echo "mosquitto_sub -t '\$SYS/#' -C 1 -p 1883 | grep -v Error || exit 1" > /healthcheck.sh
+RUN set -ex; ldconfig;
+
+#HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
+#    CMD ["/bin/sh", "-c", "/healthcheck.sh"]
+
```
iegomez, line 115: Could you explain what are these healtchecks bits that are commented?

tetofonta: That's an example for doing health checking on the process, but this is very configuration dependent so I preferred to leave it out and really leave just an hint for anyone who may require it.

For example it does not account for authentication nor tls. In this particular case the user should configure another listener bound on localhost without any authentication in order to use this particular health check command.

iegomez: Gotcha, thanks for clarifying. In that case, I'd prefer to remove it entirely or add comments explaining what it is about.

Dockerfile
```diff
@@ -1,20 +1,23 @@
...
-    mkdir -p /build/lws && \
-    tar --strip=1 -xf /tmp/lws.tar.gz -C /build/lws && \
-    rm /tmp/lws.tar.gz && \
-    cd /build/lws && \
+RUN set -ex; \
+    wget https://github.com/warmcat/libwebsockets/archive/v${LWS_VERSION}.tar.gz -O /tmp/lws.tar.gz; \
+    mkdir -p /build/lws; \
+    tar --strip=1 -xf /tmp/lws.tar.gz -C /build/lws; \
+    rm /tmp/lws.tar.gz; \
+    cd /build/lws; \
```
kfdm, line 20: Out of curiosity, why are you changing all instances of `&&` to `;` ?
I would think we would *WANT* the build to fail if any step had an error, but here it would keep going 🤔

tetofonta: It's for readibility. `set -ex` will do the same thing by exiting in case of non zero exit code of any command in the script (and will print it out before executing)

Dockerfile
tetofonta: I made a new commit removing the comments.

Do you think the argument "health checking" should be mentioned in the readme?

Dockerfile
kfdm: I'm primarily asking because I had not seen that construct before and was asking to learn myself. It's ultimately up to iegomez what he's comfortable with :)

tetofonta: Oh dang, I have forgotten line [90](https://github.com/tetofonta/mosquitto-go-auth/blob/c1c0fdda86877a78e511ca6f13dc5ed8cb069544/Dockerfile#L90) =( You are right.

As for my understanding, docker executes all the run programs as `/bin/sh -c '<run command>'` (because if you leave a RUN entry without commands it errors out with `/bin/sh: 0: -c requires an argument`)

You can see this behavior of `set -e`/`set -ex` in a linux shell:
```
/bin/sh -c "set -ex; echo 'a'; [ 1 -eq 2 ]; echo 'b'"
``` 
this will run until ` [ 1 -eq 2 ]` returns non zero so it does not execute the second echo and it is the same as
```
/bin/sh -c "echo 'a' && [ 1 -eq 2 ] && echo 'b'"
```
I'm making a commit fixing those `&&`, sorry for the mistake :innocent:

tetofonta:
As said before, I fixed the style for Dockerfile using now `set -ex` on multi commands `RUN` entries. Those `RUN` with only one command does not require this because the exit code of the block is the exit code of the command.

Dockerfile
iegomez: I'd actually leave it up to you guys, I'm not much of a Docker user and don't know what users normally expect. So if you both think the latest version of this PR is ready, I'll gladly merge it.

iegomez: So, @tetofonta, @kfdm, what do you say?

tetofonta: Sorry for the time taken to answer.
I think this is an improvement on the current Dockerfile because of the double compilation of `libwebsocket` and because of the added functionality of the mosquitto utils.

As for the style, that's kind of a personal thing and does not matter. That's as you prefer, and I can revert those modifications without too much work.

kfdm: My comment was primarily that mixing build changes and code formatting all together makes the actual changes more difficult to see. Ideally it would have been two commits,

commit 1: add `set -ex` and change `&&` to `;` without changing any logic
commit 2: change the logic to change what is built or not built

My initial comment was mostly the surprise that there was a specific reason to change `&&` to `;`

iegomez: Aa said, I don't even use the Docker setup, so though I agree 2 commits with separate changes would be nice, I'm fine with squashing and merging if you don't have the time. I don't see us reverting the change anyway.

kfdm: @tetofonta can you do a rebase of your commit and split it up into logical commits that we can more easily review? If you want to change all the `&&` to `set -ex` and `;` then those need to be their own commit, so that the actual changes can be reviewed by themselves.

iegomez: Btw, @kfdm, @tetofonta , would you be interested in being maintainers of this project?
I ask because I don't look at it much these days, every few months I get some time and start looking at issues and PRs, but I could certainly rely on a couple more hands watching it.

No pressure though, feel free to say no. And I really mean that, helping with a particular issue is great help already, I'm not trying to drag you into something you might not want.

Cheers!

kfdm: I definitely want to see this project continue (and I'm using it in my [own project](https://github.com/kfdm/django-mqtt#mosquitto-go-auth) ) though Go is not a language I use on a regular basis, and the http auth is the only one I'm familiar with, so I'm not sure if I'm one of the most qualified. I could mostly only check minor Docker or http plugin requests.

That said, for the most part this library seems to do what it does well, and may not actually need a lot of attention besides small bug fixes, unless you had some other long term plans you wanted to attempt.

iegomez: He, seems like your project is private, I got a 404.
And yeah, that's fair, the project tends to be quiet most of the time until someone brings up some issue or a feature request and depending on the time I might not be too responsive. So let's see how it goes for now, and thank you very much for your response.

kfdm: Hmm, thought I had already made it public. It should be public now.

iegomez:
Heads-up: I'll merge this tomorrow unless there's any concern.

kfdm:
Functionality it's probably all fine. All my nits were basically about formatting and git history 👍

PR #242, opened 2022-08-20
docs(README): fix wrong auth_plugin path

civilizeddev:
`go-auth.so` is not in `/etc/mosquitto/conf.d`, but `/mosquitto`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ include_dir /etc/mosquitto/conf.d
 Create some conf file (e.g., `go-auth.conf`) at your preferred location, e.g. `/etc/mosquitto/conf.d/`, and register the plugin's shared object path and desired backends with:
 
 ```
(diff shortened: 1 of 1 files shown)


iegomez:
The config file is wherever you want to set it, as the line before the change says.
I'm closing this because the change is not needed.

civilizeddev:
@iegomez

I forgot to say this.

I use mosquitto-go-auth in a k8s environment.

In your docker image:
https://hub.docker.com/r/iegomez/mosquitto-go-auth

It contains go-auth.so in `/mosquitto`

I had trouble with this because of the path of the so file.

Anyway,

I don't want anyone else to get into trouble about this.

Thank you.

iegomez:
I understand and appreciate the contribution, I think maybe we can change make it better when enhancing the docs that need some work in general anyway.

PR #248, opened 2022-10-07
Postgres tls/sslmode "verify-full" as default

maab:
With this PR `verify-full` will be the default postgres sslmode instead of `disable`. A user then has to actively disable transport layer security. This improves security and hopefully awareness of users for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,10 @@ Depending on the sslmode given, sslcert, sslkey and sslrootcert will be used. Op
 	require - Always SSL (skip verification)
 	verify-ca - Always SSL (verify that the certificate presented by the server was signed by a trusted CA)
 	verify-full - Always SSL (verify that the certification presented by the server was signed by a trusted CA and the server host name matches the one in the certificate)
+	
+From *mosquitto go auth* version 2.0.0 on `verify-full` will be the default sslmode instead of `disable`. You may have
+to disable transport layer security if the postgres database server doesn't support encryption and has a certificate
+signed by a trusted CA. 
 
 Queries work pretty much the same as in jpmen's plugin, so here's his discription (with some little changes) about them:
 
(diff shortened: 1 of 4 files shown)


iegomez (approved):

iegomez:
Thanks, @maab!

I retargeted it to `master` because `2.0.0` was out of date and I'km gonna do the release anyway.

iegomez (commented):
Oh, forgot about one thing: this needs to get documented. Could you change the Readme to be explicit about it?

Also, I'll recreate the `2.0.0` branch.

maab:
> @maab would you be available to document the change? I can throw something in the readme if not.

Sorry for the late reply, I was on vacation. I'll block time to extend the readme next week.

iegomez:
Sure, @maab, no rush. Sorry for bugging you on your vacation, hope you enjoyed your time off!

maab:
Thank you @iegomez and absolutely no problem.

I've pushed a suggestion for a extension of the readme.

iegomez:
One question, @maab: is this passing tests? I'd think this test case would need to have `ssl_mode` explicitly set to `disable` to pass: https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/postgres_test.go#L12

I'm out of the country and only have my work machine with me, which is not configured to either have working services for running tests, nor the Docker test script, so I can't reliably check.

iegomez:
@maab I tried running test through Docker but there's something weird going on in this machine and I failed to do so.
So yeah, I'll have to ask you to run them to see if the test case needs to be adjusted for this default value change.

maab:
Again, sorry for the late feedback. You're right, some of the tests failed because of the `verify-full` sslmode. I've fixed them by adding sslmode `disable` to the tests.

iegomez (approved):

iegomez:
@maab It takes me some times weeks to even respond, so no worries at all.
Thanks for checking and adjusting it, I've already merged and will release 2.0.0.

PR #285, opened 2023-06-06
Added jwt http timeout

alessandroperetti:
I added a config parameter  to set a custom JWT http response  timeout because now is hardcoded


backends/jwt_remote.go
```diff
@@ -152,7 +153,11 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions, versi
 		return nil, errors.Errorf("JWT backend error: missing remote options: %s", missingOpts)
 	}
 
-	checker.client = &h.Client{Timeout: 5 * time.Second}
+	if jwtHttpTimeout, ok := authOpts["jwt_http_timeout"]; ok && jwtTimeout {
+		clientTimeout = authOpts["jwt_http_timeout"] * time.Second
```
iegomez, line 157: This wouldn't work as `authOpts` is of type `map[string]string`.

iegomez (commented):
Besides what I commented, we'd also be lacking docs changes and tests.

iegomez (changes requested):
- There were some unrelated format changes in the README.
- The `timeout` field is misaligned in the struct definition, you should let `gofmt` take care of that.
- The test checking that the timeout is being used should go in at least one of the JWT remote tests at `jwt_test.go`.

backends/jwt_remote.go
```diff
@@ -152,7 +152,16 @@ func NewRemoteJWTChecker(authOpts map[string]string, options tokenOptions, versi
...
+	checker.timeout = 5  
+	if timeoutString, ok := authOpts["jwt_http_timeout"]; ok {
+		if timeout, err := strconv.Atoi(timeoutString); err == nil {
+			checker.timeout = timeout
+		} else {
+			log.Errorf("unable to parse timeout: %s", err)
+		}
+	}
+
+	checker.client = &h.Client{Timeout:  time.Duration(checker.timeout ) * time.Second}
```
iegomez, line 164:
```suggestion
	checker.client = &h.Client{Timeout:  time.Duration(checker.timeout) * time.Second}
```

backends/jwt_test.go
```diff
@@ -1654,3 +1654,77 @@ func TestJWTFormTextResponseServer(t *testing.T) {
...
+func TestJWTHttpDelayedResponse(t *testing.T) {
+	topic := "test/topic"
+	var acc = int64(1)
+	clientID := "test_client"
+	token, _ := jwtToken.SignedString([]byte(jwtSecret))
+	// wrongToken, _ := wrongJwtToken.SignedString([]byte(jwtSecret))
+
+	version := "2.0.0"
+
+	rightToken := token
```
iegomez, line 1667:
```suggestion
	version := "2.0.0"
	rightToken := token
```

backends/jwt_test.go
```diff
@@ -1654,3 +1654,77 @@ func TestJWTFormTextResponseServer(t *testing.T) {
...
+
+
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(5 * time.Second)
```
iegomez, line 1671: To avoid wasting time in tests, you could do the following tests:
- Set the `sleep` to just a bit over a second, e.g. `time.Sleep(1200 * time.Millisecond)`, initialize a JWT backend with a 1 second timeout and expect the check to fail due to a timeout.
- Set no `sleep` and initialize a JWT backend with a 1 second timeout and see the check succeed.

Since under the hood all checks use the same client to do the request, you can do one in e.g. user check and the other in the acl, that is move the sleep inside one of the cases below and leave the other one with no sleep.
That way you can both reuse the mock server instead of needing a second one, and you don't need to initialize your JWT backend twice since the same 1 second timeout will work to fail a long check and succeed on a quick one.

backends/jwt_test.go
```diff
@@ -1654,3 +1654,77 @@ func TestJWTFormTextResponseServer(t *testing.T) {
...
+				break
+			}
+			w.WriteHeader(http.StatusNotFound)
+		}
+
+	}))
+
+	defer mockServer.Close()
+
+
```
iegomez, line 1704:
```suggestion

```

iegomez (commented):
Just left some more tidying up and an idea for the test.
Also, please use `gofmt` before commiting.

backends/jwt_test.go
alessandroperetti: I could spend sometime in doing this PR.. Let me know.. I am unsure if `gofmt` is used well from my side. The tests are ok right now. Let me know. Thank you for your suggestions :)

backends/jwt_test.go
```diff
@@ -1654,3 +1654,89 @@ func TestJWTFormTextResponseServer(t *testing.T) {
...
 
 }
+
+func TestJWTHttpTimeout(t *testing.T) {
+
+	topic := "test/topic"
+	var acc = int64(1)
+	clientID := "test_client"
+	token, _ := jwtToken.SignedString([]byte(jwtSecret))
+	// wrongToken, _ := wrongJwtToken.SignedString([]byte(jwtSecret))
```
iegomez, line 1664: Please remove this commented line.

iegomez (approved):
There's a commented line that should be removed.
There's also some of the logic in the mock server isn't really necessary since we're just checking the timeout, but that's fine and could be used for further testing in the future.

So just remove the line and consider this approved.

Thanks again for your contribution!

backends/jwt_test.go
alessandroperetti: Done! :)

iegomez: Thanks!

iegomez (approved):

iegomez:
I ran tests locally just in case and it all went ok, so we're ready to merge.
Thanks for your work and patience, @alessandroperetti!

PR #288, opened 2023-06-11
Pass char* to Go and convert with C.GoString, avoiding unsafe use

ssiegel:
Using the char* passed by Mosquitto in a GoString struct is unsafe. The memory it points to is managed by Mosquitto, but Go will keep the pointer around for an indefinite duration, even when Mosquitto might free the memory.

By passing the actual char* to Go, we can use C.GoString to convert it, which copies the bytes into a buffer managed by Go. That way we can use it safely at any time in the future.

### DIFF
--- a/auth-plugin.c
+++ b/auth-plugin.c
@@ -38,15 +38,13 @@ int mosquitto_auth_plugin_init(void **user_data, struct mosquitto_auth_opt *auth
 
   GoInt32 opts_count = auth_opt_count;
   
-  GoString keys[auth_opt_count];
-  GoString values[auth_opt_count];
+  char *keys[auth_opt_count];
+  char *values[auth_opt_count];
   int i;
   struct mosquitto_auth_opt *o;
   for (i = 0, o = auth_opts; i < auth_opt_count; i++, o++) {
-    GoString opt_key = {o->key, strlen(o->key)};
-    GoString opt_value = {o->value, strlen(o->value)};
-    keys[i] = opt_key;
-    values[i] = opt_value;
+    keys[i] = o->key;
+    values[i] = o->value;
   }
 
   GoSlice keysSlice = {keys, auth_opt_count, auth_opt_count};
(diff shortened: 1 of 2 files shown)


ssiegel:
I encountered this issue in [mosquitto-go-auth-oauth2](https://github.com/gewv-tu-dresden/mosquitto-go-auth-oauth2/blob/master/src/main.go#L138), which keeps a cache keyed by the username. The cache is used to store access and refresh tokens acquired at login for future queries to the userinfo endpoint. With long lived sessions this cache can become corrupted. It is possible to work around it in mosquitto-go-auth-oauth2 by passing the cache keys through strings.Clone, but fixing it at the source seems to be the correct solution.

iegomez:
Thanks, @ssiegel, I'll take a look during this week.

auth-plugin.c
```diff
@@ -135,12 +128,7 @@ int mosquitto_auth_acl_check(void *userdata, const char *clientid, const char *u
     return MOSQ_ERR_ACL_DENIED;
   }
 
-  GoString go_clientid = {clientid, strlen(clientid)};
-  GoString go_username = {username, strlen(username)};
-  GoString go_topic = {topic, strlen(topic)};
-  GoInt32 go_access = access;
```
iegomez, line 141: This is the one thing that caught my eye.

Why pass `acc` directly and accept a `C.int` on `AuthAclCheck` instead of defining a `GoInt32` as we were doing before?
It's more surprising when you don't give the same treatment to `opts_count`, also defined as a `GoInt32` and passed along to `AuthPluginInit`, which receives a Go `int` instead of a `C.int`.

ssiegel: I don't think it really matters technically. I changed this in `AuthAclCheck` for mostly aesthetical reasons, to consistently have C types as parameters. `AuthPluginInit`/`opts_count` is an oversight. Would you prefer to use `C.int` there as well, or should I revert the `acc` parameter of `AuthAclCheck` to `GoInt32`?

iegomez: I don't have a preference really, was just asking if there was a particular reason to treat them differently.
Since there's not, I think we're good to merge.

Thanks!

PR #315, opened 2024-02-11
Fixed test after debian-stable is bookworm

mbv:
* updated mongo to 7.0, repo mongo contains only 7.0 for debian bookworm
 * installed redis 6.*, redis 7.* is not supported by `go-redis/redis/v8`


.dockerignore
```diff
@@ -0,0 +1 @@
+.github
```
iegomez, line 1: Why is this needed?

iegomez (commented):
Docker builds and test workflow stopped running, any idea why?

.dockerignore
mbv: This is a small optimization for local testing. I can exclude it from the PR if necessary. On line 63 in Dockerfile.runtest, we have COPY ./ ./, which copies all files. This just excludes it from the Docker context and doesn’t copy this folder to the image at this step. This means that if I change something in the .github folder, it won’t invalidate all Docker layers after COPY ./ ./.

mbv:
> Docker builds and test workflow stopped running, any idea why?

@iegomez Did you mean why it is running, e.g., on this PR?

I suppose that’s because the test.yml pipeline contains the trigger `on: push`, which means it runs on every push in any branch in THIS repo. However, this PR contains a branch from a forked repo. I think we should add another trigger:
```
pull_request:
    branches: [ master ]
```

iegomez:
@mbv nope, I really meant they're not running when they normally did. Now only CodeQL checks are running and nothing else, so I fetched and checked out the PR's branch to run tests locally. It's still building the image right now.
image

.dockerignore
iegomez: Gotcha, I was just wondering if ignoring `.github`, where workflows are declared, has anything to do with workflows no longer running.

iegomez:
@mbv sorry, I misunderstood, you're completely right. Haven't worked on this for a long while and I'm a bit rusty.

iegomez (approved):

Issue #310, opened 2024-01-24
http backend: superuser check is made for every single acl check


maglnet:
Hi, 

I'm currently using a http backend with caching and superuser check enabled and observed the following behavior:

As a superuser, for every subscribed topic the following happens:
* the cache is checked for user / topic / acc
* no cache hit
* an acl check for user / topic / acc is made
  * internally a superuser check is made -> granted
* the result of user / topic / acc is cached

As a normal user, for every subscribed topic the following happens:
* the cache is checked for user / topic / acc
* no cache hit
* an acl check for user / topic / acc is made
  * internally a superuser check (http request) is made -> denied
  * an acl check (http request) is made -> denied or granted
* the result of user / topic / acc is cached

Speaking of 50 subscribed topics, 1 superuser account and 1 normal account, this results in 150 requests.
(50 superuser checks for superuser + 50 superuser checks for normal user + 50 acl checks for normal user)

If the superuser check would be cached, this would reduce the requests to 52 requests.
 (1 superuser check per user + 50 acl checks for the normal user)

When scaling to more users and more topics this may result in an large amount of requests made for superuser and acl checks.

My configuration is like the following
```
auth_plugin /mosquitto/go-auth.so

auth_opt_backends http
auth_opt_http_host mymqttserver
auth_opt_http_port 443
auth_opt_http_getuser_uri /auth/check_user
auth_opt_http_superuser_uri /auth/check_superuser
auth_opt_http_aclcheck_uri /auth/check_acl
auth_opt_http_with_tls true
auth_opt_http_response_mode json

auth_opt_cache true
auth_opt_auth_cache_seconds 120
auth_opt_acl_cache_seconds 120
auth_opt_auth_jitter_seconds 3
auth_opt_acl_jitter_seconds 3
```

Maybe I did something wrong?
If not, is there anything that may improve this behavior / reduce requests?
I'm not sure if this happens for other remote backends, too.

Best regards,
Matthias

iegomez:
This seems odd, if an ACL check passed because the user was a `superuser` or because it wasn't but had right access, then an ACL cache record should be persisted, the cache is agnostic of the reason.

I may be wrong though, but could you check that your cache records are not actually expiring and that's why the superuser check gets conducted?

maglnet:
Hi,

I checked and my cache is stored for 20 minutes, so there are no requests within these 2 minutes, when topics get written, that's fine, but after that, a request for every written topic is made against the superuser endpoint again.

[...]

PR #319, opened 2024-02-27
Updated PBKDF2 hasher with more complex format handling

loffa:
This change adds compatibility with the hash-format used in Chirpstack 4. 

Since Chirpstack 4 is rewritten in Rust the format for storing passwords has changed a bit. The new format now follows the specification [PHC-String-Format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md).

The changes in this PR identifies if the new or old format is used in the supplied hash and then extracts parameters from that. Comparison is done only on the password part of the hash since that is equal between both old and new structure.

~~An update to Go 1.21 was needed to utilize the new built in slices package for comparison. This could be removed and rewritten for the older go version as well if preferred.~~

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/iegomez/mosquitto-go-auth
 
-go 1.21
+go 1.18
 
 require (
 	github.com/go-redis/redis/v8 v8.11.5
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/iegomez/mosquitto-go-auth
 
-go 1.18
+go 1.21
 
 require (
 	github.com/go-redis/redis/v8 v8.11.5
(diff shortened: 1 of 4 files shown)


iegomez:
Hey, @loffa!
Thanks for this, I haven't followed ChirpStack's development in years and didn't realize the format change.

I haven't reviewed your code yet, but I'm a bit hesitant on upgrading to Go 1.21 at the moment as I think that'd merit a major version bump. Just as an example, although Go is almost non existent within my team at work, we have a small service I wrote that was limited to Go 1.19 because of the client's restrictions. I can see people facing something similar at their work places.

So, if you don't mind, I'd prefer to stick to 1.18 for the time being. I'm working on improving the caching system (see https://github.com/iegomez/mosquitto-go-auth/issues/310) and I want to completely redo the whole options mechanism, aka `auth_opt_...`. When all that's done, releasing a new major version and bumping Go and dependencies would make sense.

loffa:
Hey,
I removed the need for the Go version update and implemented the slices compare function instead. The PR has been updated with these changes.

loffa:
I agree!

I can not see why it would be a security risk to log the hash-format or keys that can't be read in the hash text when they are unsupported either way.

iegomez (approved):

Issue #328, opened 2024-07-16
Allow for clientId to be passed to postgres GetUser backend password check


CompositeCoding:
Hi!

Thanks for your work on the plugin. I'm working on a refactor moving away from VerneMQ to Mosquitto for a large fleet of IOT devices. I would like to use the postgres backend, but we authenticate devices by their clientId, not username. 

Looking at the GetUser function signature, it seems feasible to pass the clientId as a parameter to the PG call so that it populates `$3`, making it usable in SQL queries. 

Please let me know what you think.

Thank you! 

https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/postgres.go#L186

CompositeCoding:
A config option to use the clientId instead of the username would also work for me

iegomez:
I'll add the option but in a general way, not tied to Postgres.
Of course, it'll be disabled by default. I'll link the PR when I'm done.

iegomez:
Scratch that, I'll allow `clientid` to be used in queries, then you can bypass `username` with something like this:
```
SELECT topic FROM acl WHERE $1 = $1 AND rw = $2 AND user_name_client_id_field = $3
```

iegomez:
I went with the original idea because it was simpler, the second one would involve adding support everywhere and changing the interfaces.
Here's the PR: https://github.com/iegomez/mosquitto-go-auth/pull/329

iegomez:
@CompositeCoding sorry for the time it took, thereś something going on with TLS Postgres tests, but it's unrelated to this issue and the PR I opened, so I finally decided to merge and I'll open an issue to address that.

It's in main branch now, I'll make an official release after tackling the PG tests issue and https://github.com/iegomez/mosquitto-go-auth/issues/332 that should be pretty easy to add to the release.

PR #329, opened 2024-07-21
[issue-328]: Add option to use clientid instead of username when running checks.

iegomez:
As requested in https://github.com/iegomez/mosquitto-go-auth/issues/328, this PR adds an option to override `username` with `clientid` when running any check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,19 @@ auth_opt_backends files, postgres, jwt
 
 Set all other plugin options below in the same file.
 
+#### Using clientid as username
+
+You may choose to override chedk against `username` to be done against `clientid` by setting this option:
+
+```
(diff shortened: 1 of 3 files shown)


iegomez:
Mm... this same Docker test script is passing fine in my machine, not sure yet what's going on.

iegomez:
@maab I know you did https://github.com/iegomez/mosquitto-go-auth/pull/248 a long time ago, but now TLS related tests are failing.
Do you know what could be causing the failures? Maybe Postgres version?

iegomez:
I tried downgrading PG version and it still fails, I'm not sure those test ever worked really.
That said, I tried on main branch and it failed, so it's unrelated to this PR and thus I'll merge and figure that out later.

Issue #332, opened 2024-08-29
HTTP super user route called even if ACL yet defined in file backend


Bacto:
Hi,

First I would like to thank you for your very useful authentication system!

I'm using it with 2 backends: `file` and `http`.

I have a user created in the `file` backend.
I added an equivalent of super user to this user using the following ACL file rule:
```
user myUser
topic readwrite #
```

When this user connects,  the`super user` route on the `http` backend is always called.

I think this is not the expected behavior as the user is yet a super user.

Is there a way to avoid this situation?
My problem is that it overloads the `http` backend unnecessarily.

Thanks,
Adrien

iegomez:
Sadly, unless you disable super user checks for http backend or globally, you'll get this behavior because checking an ACL first loops through superuser enabled backends to check for it, and then loops through backends to check the ACL.

I could add an option to instead of checking superuser for all backends and then acl for them, instead we loop through backends and check superuser and acl on them. That way, if `files` was registered before `http` you'd get what you expect.

But that's not implemented right now, so the short answer is the first comment: only way is to disable super user checks for http or generally.

iegomez:
Inverting the loop logic on a flag should be simple enough to implement, the only caveat is that I need to introduce ordering to backends as well.
I'll let you know when I take a stab at it.

PR #336, opened 2024-09-19
[issue-332]: Allow backend checks to be exhausted in order

iegomez:
@Bacto, this PR tackles https://github.com/iegomez/mosquitto-go-auth/issues/332.

In particular, adds the option `exhaust_backend_first` to enable checking backends for superuser and ACL in order instead of checking them all for superuser first.


iegomez:
@bacto I'll assume you're fine with the change so I'll merge.

PR #337, opened 2024-09-20
if `auth_opt_jwt_skip_user_expiration` enabled, in case of receive bad token the code crashes. also README descriptions added.

mahdizadsar:
**issue 1**:  In README.md ACL in MySQL local mode configuration had some problems, 
**Workaround**:  I realized that it should be explained to the user what parts are required to use the mysql database in JWT mode. Also to use ACL In JWT mode, `auth_opt_jwt_mysql_aclquery` must be set with the correct query. this fields are:
`auth_opt_jwt_mysql_dbname` , `auth_opt_jwt_mysql_user` ,  `auth_opt_jwt_mysql_password` , `auth_opt_jwt_mysql_aclquery`   

**issue 2**:  When `auth_opt_jwt_skip_user_expiration` is enabled in config file and the wrong JWT token is sent by client to server (with a few or completely wrong segments), the code crashes.
**Workaround**:  modify the code structure by moving the checking of token expiration conditions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1000,11 +1000,14 @@ auth_opt_jwt_userquery select count(*) from test_user where username = $1 limit 
 Thus, the following specific JWT local options are supported:
 
 
-| Option                  | default   | Mandatory | Meaning                                                  |
-| ----------------------- | --------- | :-------: | -------------------------------------------------------- |
-| auth_opt_jwt_db         |  postgres |     N     | The DB backend to be used, either `postgres` or `mysql`  |
-| auth_opt_jwt_userquery  |           |     Y     | SQL query for users                                      |
-
+| Option                        | default   | Mandatory | Meaning                                                  |
+| ----------------------------- | --------- | :-------: | -------------------------------------------------------- |
+| auth_opt_jwt_db               |  postgres |     N     | The DB backend to be used, either `postgres` or `mysql`  |
+| auth_opt_jwt_userquery        |           |     Y     | SQL query for users                                      |
+| auth_opt_jwt_mysql_dbname     |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_user       |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_password   |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_aclquery   |           |     Y/N   | ACL query must set if auth_opt_jwt_db set is `mysql`     |
 
 Notice that general `jwt_secret` is mandatory when using this mode.
 `jwt_userfield` is still optional and serves as a mean to extract the username from either the claim's `Subject` (`sub` field),
@@ -1022,7 +1025,7 @@ auth_opt_jwt_userquery select count(*) from "user" where username = $1 and is_ac
 For mysql:
 
 ```
-auth_opt_jwt_userquery select count(*) from "user" where username = ? and is_active = true limit 1
+auth_opt_jwt_mysql_aclquery select count(*) from "user" where username = ? and is_active = true limit 1
 ```
 
 *Important note:*
--- a/README.md
+++ b/README.md
@@ -984,12 +984,16 @@ initMqttClient(applicationID, mode, devEUI) {
 #### Local mode
 
 When set to `local` mode, the backend will try to validate JWT tokens against a DB backend, either `postgres` or `mysql`, given by the `jwt_db option`.
-Options for the DB connection are the almost the same as the ones given in the Postgres and Mysql backends but prefixed with `jwt_`, e.g.:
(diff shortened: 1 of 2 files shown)


backends/jwt.go
```diff
@@ -132,15 +132,19 @@ func getJWTClaims(secret string, tokenStr string, skipExpiration bool) (*jwtGo.M
...
+			if skipExpiration {
+				expirationError = true
+			}else{
+				log.Debugf("jwt parse error: %s", err)
+				return nil, err	
+			}
+		}else{
+			log.Debugf("jwt parse error: %s", err)
+			return nil, err	
+		}
```
iegomez, line 147: The format is a bit weird, Go's formatter should have caught missing spaces.

Also, I'd structure it like this instead to skip the inner `else` clause by returning early on `!skipExpiration`:
```
	if err != nil {
		if v, ok := err.(*jwtGo.ValidationError); ok && v.Errors == jwtGo.ValidationErrorExpired {
			log.Debugf("jwt token expired: %s", err)

			if !skipExpiration {
				return nil, err
			}

			expirationError = true
		} else {
			log.Debugf("jwt parse error: %s", err)

			return nil, err
		}
	}
```

iegomez: On another note, I'd also add a test with some wrong token that previously would crash and now doesn't.

iegomez (changes requested):

README.md
```diff
@@ -1000,11 +1000,14 @@ auth_opt_jwt_userquery select count(*) from test_user where username = $1 limit
...
-| auth_opt_jwt_userquery  |           |     Y     | SQL query for users                                      |
-
+| Option                        | default   | Mandatory | Meaning                                                  |
+| ----------------------------- | --------- | :-------: | -------------------------------------------------------- |
+| auth_opt_jwt_db               |  postgres |     N     | The DB backend to be used, either `postgres` or `mysql`  |
+| auth_opt_jwt_userquery        |           |     Y     | SQL query for users                                      |
+| auth_opt_jwt_mysql_dbname     |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_user       |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_password   |           |     Y/N   | must set if auth_opt_jwt_db set is `mysql`               |
+| auth_opt_jwt_mysql_aclquery   |           |     Y/N   | ACL query must set if auth_opt_jwt_db set is `mysql`     |
```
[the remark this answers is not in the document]
mahdizadsar: I agree, but I suggest that the tables should be completed in order to explicitly specify which items are mandatory for a particular situation. In this case, to use MySQL in JWT mode, the fields I wrote seem mandatory

README.md
```diff
@@ -1022,7 +1025,7 @@ auth_opt_jwt_userquery select count(*) from "user" where username = $1 and is_ac
 For mysql:
 
 ```
-auth_opt_jwt_userquery select count(*) from "user" where username = ? and is_active = true limit 1
+auth_opt_jwt_mysql_aclquery select count(*) from "user" where username = ? and is_active = true limit 1
```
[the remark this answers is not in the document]
mahdizadsar: right, but you wrote an example for PG in a few lines earlier. This line is an example for MySQL, however, this is similar to the previous topic.

backends/jwt.go
mahdizadsar: great good job

README.md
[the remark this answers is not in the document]
iegomez: Yeah, I'm not against the addition but the replacement.

README.md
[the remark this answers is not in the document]
iegomez: Sure, I was just saying that besides the user query, which becomes irrelevant, all the options still follow the same _mandatoriness_ that's spelled on PG and MySQL sections.
That said, I'm not really against being super clear, so if you want to throw in tables for both DBs in the JWT case, all the better.

backends/jwt.go
iegomez: Written comments don't convey tone so I'll assume good will, so all I'm gonna say is that if you address my comments, I'll merge.

iegomez: @mahdizadsar let me know if you plan on adding that test and addressing other concerns so we may merge.

mahdizadsar: @iegomez 

Based on your intention, I changed a few things:
In the **README.md** file:
Changes in the local mode section in the JWT sub-set:
1- I linked the words **Postgres** and **Mysql** to their own section.
2- added an example for Mysql similar to Postgres.
3- reverted the options table to its previous state.
4- reverted the changes related to the _userquery_ section to its previous state.
6- In the Important note section, I put the keywords in back ticks.
5- I added only one example for connecting to the **Mysql** database at the end.

In the **jwt.go** file:
I changed the file similar to what you wrote so that the "_if else_" condition becomes more readable as you said.

mahdizadsar (commented):
@iegomez 
I reviewed the changes and added an improvement to the README.md file.
Of course, if you agree with me.

PR #338, opened 2024-09-22
feat: add mongo_uri config option

rickbassham:
Adds a new `mongo_uri ` config option to specify the entire mongo uri. This allows for more complex connection strings.

### DIFF
--- a/backends/mongo.go
+++ b/backends/mongo.go
@@ -108,7 +108,13 @@ func NewMongo(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
 		m.insecureSkipVerify = true
 	}
 
-	addr := fmt.Sprintf("mongodb://%s:%s", m.Host, m.Port)
+	var addr string
+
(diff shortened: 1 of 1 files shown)


iegomez (commented):
Could you add a test for the change?

backends/mongo.go
```diff
@@ -108,7 +108,13 @@ func NewMongo(authOpts map[string]string, logLevel log.Level, hasher hashing.Has
 		m.insecureSkipVerify = true
 	}
 
-	addr := fmt.Sprintf("mongodb://%s:%s", m.Host, m.Port)
+	var addr string
+
+	if mongoUri, ok := authOpts["mongo_uri"]; ok {
+		addr = mongoUri
```
iegomez, line 114: Is there some standard validation to run against the `uri`?

rickbassham: The existing `mongo.Connect` method will validate it and return an appropriate error if the URI is bad.

iegomez:
@rickbassham let me know if you plan on adding that test I mentioned.

PR #340, opened 2024-10-15
feat: add-host-prefix-option

joda01:
Allow to configure the host prefix. This allows to use mongodb and mongodb-srv


iegomez (commented):
Any way you can add some test for this?

Issue #189, opened 2021-08-26
Enhancement : LDAP


g0lgs:
I would like to see LDAP support added.

iegomez:
Well, the "help wanted" label didn't help at all, so I'll sadly close this.

PR #351, opened 2025-05-07
LDAP Backend

### LINKED ISSUE #150: Feature Request: LDAP/AD
It would be great if LDAP was a supported method. I'd hope that you could just update / translate into Go [jpmens/mosquitto-auth-plug LDAP code](https://github.com/jpmens/mosquitto-auth-plug/blob/master/be-ldap.c).

It looks like Go has a good LDAP library [go-ldap](https://godoc.org/gopkg.in/ldap.v2) that would make it relatively painless. They even have an [example for user auth](https://godoc.org/gopkg.in/ldap.v2#ex-package--UserAuthentication).

lhns:
Hi, I implemented an LDAP Backend to fix #150 and #189.


lhns:
For anyone who wants to test this, you can try this docker image: ghcr.io/lhns/mosquitto-go-auth:ldap
It has been working pretty well on my setup for a few days now.

iegomez:
Hi, sorry I didn't look into this earlier, my personal machine was busted and I prefer not to use the work one for personal stuff.
I'm finishing configuring a new one and will take look once I'm done.

lhns:
TestGoStore seems to be flaky even if i increase all the delays by 5 times...

iegomez:
Yep, it failed on the checks run in the PR.
Postgres is failing as well, but that was introduced on a version change some time ago and is unrelated, while the bump on Go version in this PR might be making 'GoStore' fail: I took a look and that original key/value in memory store hasn't seen an update in 6 years, so might be it.

I'll check https://github.com/jellydator/ttlcache to see if it's worth a switch.

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+
+	l, err := NewLDAPWithFactory(authOpts, logLevel, func(l LDAP) (LDAPClient, error) {
+		ldapClient, err := ldap.DialURL(l.Url)
+		return ldapClient, err
+	})
+
+	return l, err
+}
+
+func NewLDAPWithFactory(authOpts map[string]string, logLevel log.Level, ldapClientFactory LDAPClientFactory) (LDAP, error) {
```
iegomez, line 45: I understand the desire of decoupling the configuration of the backend with the actual connection to it given your tests mocks them, and I'm not a priori against it, but I must ask:
- Are you coming from Java Factory world (just kidding)?
- Isn't there a way to test the backend against a dev ldap server like ldap-utils or something?

I know it's a hassle to setup a real ldap service, but Postgres' failures have shown that if you don't test against the real thing, you might be fooling yourself because the mock works, but the actual service doesn't.

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+		ldap.NeverDerefAliases,
+		0,
+		0,
+		false,
+		fmt.Sprintf(l.UserFilter, ldap.EscapeFilter(username)),
+		[]string{"dn"},
+		nil,
+	)
+
+	searchResult, err := l.LDAPClient.Search(searchRequest)
```
iegomez, line 150: [Nitpick] Could you add empty lines between statements (not only here, but in general)?

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+
+type LDAPClient interface {
+	Close() error
+	Bind(username, password string) error
+	Search(request *ldap.SearchRequest) (*ldap.SearchResult, error)
+}
+
+type LDAP struct {
+	LDAPClientFactory        LDAPClientFactory
+	LDAPClient               LDAPClient
```
iegomez, line 22: Already commented on the `Factory` thing, but apart of that the struct is already named `LDAP`, its client doesn't need the prefix as well since it's private, it may simply be called called `client`.

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+
+	return true, nil
+}
+
+func (l LDAP) GetSuperuser(username string) (bool, error) {
+
+	//If there's no superuser filter, assume all privileges for all users.
+	if l.SuperuserFilter == "" {
+		return false, nil
+	}
```
iegomez, line 188: Maybe I'm reading this wrong, but isn't it returning the contrary that the comment says?
I.e. when there's no filter set, `GetSuperuser` will always return false.

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+	if len(searchResult.Entries) != 1 {
+		log.Debugf("LDAP superuser search returned %d entries", len(searchResult.Entries))
+		return false, err
+	}
+
+	return true, nil
+}
+
+// checks whether an MQTT topic matches a pattern with wildcards (+, #) according to MQTT spec rules.
+func matchMQTT(topic, pattern string) bool {
```
iegomez, line 216: Is there a reason this backend needs a specialized check that's not covered by https://github.com/iegomez/mosquitto-go-auth/blob/master/backends/topics/topics.go?

If there's a bug in that function, then I'd gladly welcome a fix for it; otherwise, backends should not have their own topic checking.

backends/ldap.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	//If there are no acl attributes defined, assume all privileges for all users.
+	if len(attributes) == 0 {
+		return true, nil
+	}
+
+	//If there is no groupBaseDN, return false.
+	if l.GroupBaseDN == "" {
+		log.Debugf("ldap_group_base_dn not set, cannot check ACL")
```
iegomez, line 268: Here is an example, but the comment is in general: if something's an error, I'd log it as such, not at the `debug` level.

iegomez (commented):
Just left a few comments, tomorrow I'll do a more thorough review.

backends/ldap.go
lhns: Oh thats really nice, I must have overlooked it. I will add the tests I already have for that.

backends/ldap.go
lhns: Good point!

backends/ldap.go
lhns: Well, actually I have quite a few java projects XD

I didn't find an easy ldap server to do this with yet. Many ldap are a real hassle to set up in tests. Is ldap-utils a library? I cannot find it. Your point is valid though.

backends/ldap.go
lhns: must have slipped in a refactoring

backends/ldap.go
lhns: Sure thing

backends/ldap.go
lhns: Allthough, the more I think about it, the LDAP Factory would bring the possibility to do connection pooling in the future.

lhns: I think I found a few ldap implementations that would be easy to integrate into the test setup

lhns: https://github.com/jimlambrt/gldap/tree/main/testdirectory looks pretty nice. I will update the tests

iegomez: Cool, thanks!

lhns: Well it turns out that testdirectory doesn't support stuff like memberOf and a few other things. I've decided to use proper openldap, which was really painful to configure but I its finally done!!

iegomez:
> @iegomez could you re-review?

Yeah, sorry, I was out sick yesterday.
I'll recheck today.

iegomez (approved):
I think I caught one repeated test, but otherwise, great work!

lhns:
Thanks! I removed the superfluous test.

PR #353, opened 2025-05-25
Add tests to topics.Match and constrain # matcher

lhns:


iegomez:
I've skipped Postgres TLS tests until I fix them, could you rebase from master?

iegomez (approved):

